PR #48, opened 2017-06-14
Rename 06 and add subheadings

Robinlovelace:


Robinlovelace:
Yes I considered the downsides including this one hence putting it in a PR. How about "Reprojecting and Transforming Geographic data" or just "Reprojections and Transformations"? Note we will say something in Chapter 2 on CRSs and setting this that can cross-reference this chapter.

I think CRSs don't merit an entire chapter and that CRS transforms are just an extension of affine transforms so they will fit in the same chapter nicely.

Nowosad:
Yes! I like this chapter idea. Maybe "Reprojections and Transformations" for now, and we could change it later?

PR #49, opened 2017-06-23
13-gis

jannes-m:
The chapter still requires some work so you don't have to merge the request right now. This way you can easily keep track of what I am doing.


Robinlovelace:
Looking at this now: great work.

13-gis.Rmd
```diff
@@ -133,28 +133,148 @@ In this case the output tells us, had we left the `FIELDS`-parameter unspecified
...
+Specifically, we would like to compute the SAGA wetness index from a digital elevation model.
+First of all, we need to make sure that **RSAGA** will find SAGA on the computer when called.
+For this, all **RSAGA** functions using SAGA in the background make use of `rsaga.env()`. 
+Usually, `rsaga.env()` will detect SAGA automatically by searching several likely directories (see its help for more information). However, we have 'hidden' SAGA in the OSGEO4W-installation, a location `rsaga.env()` does not search automatically. Therefore, to ensure that `rsaga.env()` finds the SAGA installation, we add the corresponding path to the PATH environment variable. Naturally, these steps are unnecessary under Linux and MacOS. 
 
-hint to vignette("RSAGA-landslides", package = "RSAGA").
+```{r}
+library("RSAGA")
+path <- Sys.getenv("PATH")
+path <- paste(c("C:\\OSGeo4W64\\apps\\saga", path), collapse = ";")
```
Robinlovelace: This chunk should probably have `eval=FALSE` if it uses hard-coded paths. Thoughts @Nowosad?

jannes-m: Well, we could use link2GI to make it reproducible on other system setups as well. Good point. I will take care of this. However, this is only necessary under Windows.

Robinlovelace: Awesome level of integration and consistency

Nowosad: I'll take a look on this in the afternoon.

jannes-m: I have already updated this, see 06966e5. And since there is a tiny slash problem, I have opened a new link2GI issue (see https://github.com/gisma/link2GI/issues/3)

Robinlovelace:
Hi @Nowosad these are good points thanks for that and great to have a second pair of eyes on this. Only comments: YES please use our style (which type of English to be decided soon) and yes the packages mentioned should be added but only *need* to be added when we start building this thing on Travis which may be slightly easier said than done.

jannes-m:
Hi Jakub,
thanks for your comments! 
1.Since you still seem to discuss the assignment operator, I just want to share my view.Though the discussion what to use at times seems a bit esoteric,I like the clear distinguition between assigning to an object (<-) and assigning to a function parameter (=). But if you opt for =, no problem :)!
2. Well, you never know with QGIS :) But set_env will find out if you are using QGIS3. And then I hope we only have to provide a second python_funs.py, and everything else will stay the same.At least I have tried to construct it this way,i.e.I tried to avoid substantial Python programming in the R processing part.If it works, we'll find out.Anyways, I will come up with a solution, no worries.
3.Agreed.Will provide them when I am done with rgrass7
4.To be honest,I was aware of the package but never used it.Does it provide some additional functionality that you don't have with RQGIS? In any case, we should definitely mention it.

Nowosad:
@jannes-m 
Ad 1. We had that <- vs = discussion before and decided on using =. One very important thing about style - each sentence (not each paragraph) should start with a new line. In this way it is easy to see changes in the text. Take a look at the source code of the first few chapters.
Ad 4. I use gdalUtils, when I need to do some fast raster conversions, such as to crop a file. I'm not sure, but I always thought that the raster functions in QGIS are in fact gdal functions. Can you clarify that?

jannes-m:
Ok, then I'll change everything from <- to =. And yes, I always try to use one line per sentence. Admittedly, here I have forgotten to do so, sorry. But I had already changed that.

QGIS uses GDAL for data import/export. And in the processing toolbox you also have access to GDAL functions:

    library("RQGIS")
    find_algorithms("gdalogr:")
    # and more specifically
    find_algorithms("gdalogr:(.*)?clip")

That's why I wondered if it was necessary to explain gdalUtils in detail. That would make sense if this package would provide additional functionality.
As far as I know there are no native QGIS functions performing clip-operations on rasters. But of course you could use GRASS, SAGA or other third-party providers for that.

Nowosad:
@jannes-m I've checked right now. In QGIS: Raster->Extraction->Clipper (and probably most, if not all, of raster functions) use GDAL. In case of Clipper it's gdal_translate.

Robinlovelace:
This chapter is looking better and better. No rush to get on Travis but would be good to merge soon-ish. What are your thoughts on this @Nowosad?

Nowosad:
I'm fine with that.

Robinlovelace:
Many thanks Jannes for the chapter.

jannes-m:
@Nowosad Jakub, please check the QGIS processing toolbox in the GUI to find out which raster algorithms are provided by QGIS. These are only five algorithms (Raster general tools & Raster tools). And now we would have to look at the source code to find out if these are really native QGIS algorithms or if they also use other software in the background. In the processing toolbox you can also find all GDAL/OGR algorithms (in my case 45). And this should correspond more or less to:
    
    library("RQGIS")
    find_algorithms("gdalogr:")

But of course, you can also make use of all the raster functionality provided by GRASS and SAGA, again check out the processing toolbox. The menu (e.g., raster - extraction - ...) only provides a subset of all available raster geoalgorithms.

@Robinlovelace sure, it was fun writing this chapter. I will finish it when I come back from Iceland (1.08.).

PR #63, opened 2017-08-02
Sf class plot

jannes-m:
Two or three weeks ago I have been working on a figure representing the sf-classes. If you like the figure, merge it, otherwise simply reject the pull request. I just didn't want to put this figure stuff with 13-gis stuff into one pull request.


Robinlovelace:
Thanks Jannes. Just to confirm does this result in the plot pasted below? If so I reckon we can do better than that - @Nowosad has a version working that actually uses the geom types as part of the plot which is pretty neat.

sf-classes-jannes

jannes-m:
Yes, exactly, this is the plot. And yes, Jakub's plot was much prettier. However, the plot was only intended as a conceptual model since Jakub asked if it was a good idea to put the geometry collection in the middle of the plot (46f50e4). But I see, you have decided to go for Jakub's version for the first review round. Ok, then dismiss this pull request.

Robinlovelace:
It's useful to conceptualise it in this way but I think easier to understand with the plots @Nowosad is working on. Closing for now but could be useful to refer readers to or to revive if reviewers think this is a better representation.

PR #64, opened 2017-08-02
13-gis

jannes-m:
The last two days I have been working on the GRASS subsection. It took me a bit longer since I haven't used rgrass7 that much in the past. However, I am quite impressed of this powerful, and user-friendly interface. But one has to be familiar with GRASS to harness its real power. To finish the chapter I only have to refine the concluding section "What to use when". I will take care of this within the next ten days (now I will look into the introduction). Until then you might just leave the pull request open.


Robinlovelace:
Looks good, not tested but I like what I've read. @Nowosad is better placed than me to check-over this: does it build OK if we add chapter 13? You could add uncomment the chapter here to check: https://github.com/Robinlovelace/geocompr/blob/master/_bookdown.yml

Nowosad:
I am going to try it tomorrow.

Nowosad:
Ok, I've tested it on one of my own laptops. I was unable to reproduce several parts of it, but it could be because I've very unusual configuration on this machine. I'm going to test it one more time on my work computer on Monday.

jannes-m:
If you have installed SAGA 2.3.1, the SAGA section will not work. When writing the SAGA subsection, SAGA 2.3.1 has not been the LTR. RSAGA is currently not supporting SAGA 2.3.1 but we have already filed an issue (https://github.com/debangs/RSAGA/issues/16). So we will have to wait until RSAGA has been updated. If they don't do it, we will have to do it ourselves...
By contrast, the RQGIS section should work, and also the GRASS7-subsection. Please use the RQGIS developer version (`devtools::install_github("jannes-m/RQGIS")`) before running these sections since there have been some issues since the release of QGIS 2.14.17 and 2.18.11.

PR #66, opened 2017-08-04
history additions

rsbivand:
Jakub asked for comments, here are a few suggestions.


Nowosad:
@rsbivand Thank you.

Robinlovelace:
Just read over these. Thanks for highlighting maptools and its class `Map` - could also add that it has a wealth of functions still useful today like `SpatialLinesMidPoints()` (which I've used in my own work) that work on **sp** classes. The history of **spdep** is also an interesting and important addition many people will benefit from knowing.

PR #74, opened 2017-08-16
reviewing chapter 2

jannes-m:
Just started with it


Robinlovelace:
@Nowosad I suggest we merge this if you're happy with this to keep things rolling. Thanks @jannes-m for the improvements - great if we can iterate on this quick that will be great, although I'm heading off for a 'long weekend'.

Nowosad:
@Robinlovelace I'm going to read the changes and merge it tomorrow morning.

02-spatial-data.Rmd
```diff
@@ -802,7 +796,7 @@ st_crs(sf_points)
 ```
 
 ## Units
-
+<!-- If this is the final thing about sf, we should move it up before starting to explain raster stuff -->
```
Nowosad: @Robinlovelace what's your opinion?

02-spatial-data.Rmd
```diff
@@ -212,29 +224,33 @@ For more advanced map making we recommend using a dedicated visualization packag
...
 
 ```{r, eval=FALSE, echo=FALSE}
 plot(africa[0], lwd = 3, main = "Nigeria in context", border = "lightgrey")
 ```
 
 #### Further work
 
-**sf** makes R data objects more closely aligned to the data model used in GDAL and GEOS, in theory making spatial data operations faster.
-The work here provides a taste of the way **sf** operates but there is much more to learn (see Chapter \@ref(spatial-data-operations)). And there is also a wealth of information in the highly recommended vignettes of the package.
+**sf** makes R data objects more closely aligned to the data model used in GDAL and GEOS, in theory making spatial data operations faster <!--I am not sure what you mean-->.
```
Robinlovelace: Fair point - the meaning is hidden. I suggest:

```
**sf** simplifies spatial data objects compared with **sp** and provides a near-direct interface to GDAL and GEOS C++ functions.
In theory this should make **sf** faster than **sp**/**rgdal**/**rgeos**, something that is tested in Chapter 5 for data I/O.
This section introduces **sf** classes in preparation for subsequent chapters which deal with vector data (in particular Chapter \@ref(spatial-data-operations)).
```

I've removed the sentence about vignettes. I believe we mention them elsewhere (there are now 4).

02-spatial-data.Rmd
```diff
@@ -253,36 +269,38 @@ plot(world_centroids, add = TRUE, cex = world$pop / 1e8, lwd = 3)
...
+Generally, well-known binary (WKB) or well-known text (WKT) are the standard encoding for simple feature geometries.
+WKB representations are usually hexadecimal strings easible readable for computers.
+This is why, GIS and spatial databases use WKB to transfer and store geometry objects.
+WKT, on the other hand, is a human-readable text markup description of simple features. 
+Both formats are exchangeable, and if we present one, we will naturally choose the WKT representation.
+
+The basis for each of the geometry types is the point. 
+A point is simply a coordinate in 2D, 3D or 4D space:
+
+- X and Y coordinates represent the location of the points on the surface of the Earth. Depending on the coordinate reference system you use, this can be, for example, easting and northing or longitude and latitude.
```
Robinlovelace: Good point. In fact **sf** is *not good* at dealing with data in geographic (on the surface of the Earth) CRSs because GEOS itself assumes a planar projection. The overall direction is good though. I suggest:

```
- X and Y coordinates represent the location points in 2D space, with units depending on the coordinate reference system (CRS) of the geometry data.
Commonly used units include easting and northing in meters or longitude and latitude (see Chapter 6). 
```

We can update the cross-reference when chapter 6 goes 'live'!

02-spatial-data.Rmd
Robinlovelace: I was thinking we could talk about units in a way that is relevant to raster and vector data. With raster data the most important thing is the resolution and (I don't think) the **raster** package has native support for units. I think that's worth mentioning at the end of the chapter. So I suggest:

1. An introductory paragraph on the importance of units for geographic data.
2. A few paragraphs explaining units in **sf**, e.g. beginning with a sentence like `An innovative feature of **sf** compared with its predecessor **sp** is that it has *native support* for units.` followed by some worked examples.
3. A paragraph or 2 on units in raster data with a worked example.

jannes-m:
Sure, as soon as we have agreed on the pending review issues!

02-spatial-data.Rmd
```diff
@@ -24,24 +28,29 @@ library(raster)
...
 ## Vector data
 
 Vector data are based on points that are located on a Cartesian (or geographic) coordinate system.
-Each point in vector data is typically described by two numbers representing distance from the $origin$ along the $x$ (horizontal) and $y$ (vertical) axis in Euclidean space.^[In 3 dimensional coordinate systems three numbers are needed for each of the three axes, $x$, $y$ and $z$.]
+Typically, a point consists of an x and a y value. 
+These two numbers simply represent the distance from the $origin$ along the $x$ (horizontal) and $y$ (vertical) axis in Euclidean space.^[In 3 dimensional coordinate systems three numbers are needed for each of the three axes, $x$, $y$ and $z$.]
 In mathematical notation these points are typically represented as numbers separated by commas and enclosed by a pair of brackets: 
 $(1, 3)$ for example, represents a point located one unit to the right and three units above the origin.
-There is clear link between these vector points and the `vector` class in R.
+There is a clear link between these vector points and the `vector` class in R.
```
Robinlovelace: Agreed - the previous phrasing risks confusing readers. There is a clear distinction between an abstract data model and a data class implemented in a language such as R.

02-spatial-data.Rmd
```diff
@@ -212,29 +224,33 @@ For more advanced map making we recommend using a dedicated visualization packag
 - map export 
 -->
 
-<!-- Maybe show also somewhere that `world[0]` produces only a plot of the geometry which is rather useful if you do not want to plot a specific attribute. This way, you can for example dismiss the col = "white"-argument in your Nigeria example. -->
+<!-- Maybe show also somewhere that `world[0]` produces only a plot of the geometry which is rather useful if you do not want to plot a specific attribute. This way, you can for example dismiss the col = "white"-argument in your Nigeria example.
```
Robinlovelace: Yes - please do mention the `plot(x[0])` behaviour in the prose: interesting and potentially useful to people who like ultra-concise code.

Robinlovelace: Then I think this line can be deleted.

02-spatial-data.Rmd
```diff
@@ -212,29 +224,33 @@ For more advanced map making we recommend using a dedicated visualization packag
...
+First, we convert the countries of the world into centroids, and then subset those in Asia. Finally, the `summary`-command tells us how many centroids (countries) are part of Asia (43) and how many are not (134).
 
 ```{r}
 world_centroids = st_centroid(world)
 sel_asia = st_intersects(world_centroids, asia, sparse = FALSE)
 summary(sel_asia)
 ```
 
-Note: another way of achieving the same result is with a GEOS function for identifying spatial overlay, which we'll cover in more detail in Chapter \@ref(spatial-data-operations).
+<!--Note: another way of achieving the same result is with a GEOS function for identifying spatial overlay, which we will cover in more detail in Chapter \@ref(spatial-data-operations).-->
```
Robinlovelace: Fair enough. Do you have a suggestion on the ideal solution for that? Feel free to update the content to represent that ideal world - I think it's useful to say that sometimes you can achieve the same result using a spatial operation (using GEOS in the background as you say - happy to spell that out or just not mention GEOS as you prefer) *and* an attribute operation.

02-spatial-data.Rmd
```diff
@@ -253,36 +269,38 @@ plot(world_centroids, add = TRUE, cex = world$pop / 1e8, lwd = 3)
...
+In this chapter we will focus on the seven most commonly used types: `POINT`, `LINESTRING`, `POLYGON`, `MULTIPOINT`, `MULTILINESTRING`, `MULTIPOLYGON` and `GEOMETRYCOLLECTION`.<!--FIG-->
+Find the whole list of possible feature types in [the PostGIS manual ](http://postgis.net/docs/using_postgis_dbmanagement.html).
+
+Generally, well-known binary (WKB) or well-known text (WKT) are the standard encoding for simple feature geometries.
+WKB representations are usually hexadecimal strings easible readable for computers.
+This is why, GIS and spatial databases use WKB to transfer and store geometry objects.
+WKT, on the other hand, is a human-readable text markup description of simple features. 
+Both formats are exchangeable, and if we present one, we will naturally choose the WKT representation.
+
+The basis for each of the geometry types is the point.
```
Robinlovelace: I didn't write this but agree that it's not the most enlightening content of this chapter and suggest we can delete this for now. It also seems to duplicate stuff at the beginning of the chapter where we plot a point in Euclidean space. In summary, agree: delete (heads-up @Nowosad we can always re-add this).

02-spatial-data.Rmd
```diff
@@ -296,29 +314,28 @@ They could be easily describes as a well-known text:
...
-A linestring is represented by a sequence of points with linear interpolation between points, for example:
+A linestring is a sequence of points with a straight line connecting the points, for example:
 
 - `LINESTRING (1 5, 4 4, 4 1, 2 2, 3 2)`
 
 ```{r linestring, echo=FALSE}
 plot(st_as_sfc("LINESTRING(1 5, 4 4, 4 1, 2 2, 3 2)"), axes = TRUE, main = "LINESTRING")
 ```
 
-Linestring cannot have self intersecting line part.
```
Robinlovelace: Again not my work but agree with the comment and suggest the self-intersecting part to keep it simple. Please make the necessary change to do this.

02-spatial-data.Rmd
```diff
@@ -611,14 +622,12 @@ Simple features are, in essence, data frames with a spatial extension.
 
 ### An introduction to raster
 
-Raster objects in R are supported by the **raster** package. 
-It provides an extensive set of functions to create, read, processed and write of raster datasets.
-Beside the general raster data manipulation, **raster** provides many low level functions that can be used to create and develop new concepts. **raster** also supports work on large raster datasets that are stored on a hard drive, but are too large to fit into memory. 
-Instead of recreating the whole file in RAM, this package extracts information about the structure of the dataset, such as a number of rows and columns, spatial extent and the name of the file.
-When manipulating this dataset, values are read and processed in a small chunk and written either to a specified file on a disk or temporary file.
-
-The **raster** package provides three main classes of objects - `RasterLayer`, `RasterBrick` and `RasterStack`.
```
Robinlovelace: Agree - duplication of content is a problem in the book we need to keep on top of. Fresh eyes are very much appreciated in this area and I agree these sentences should be deleted. Good change.

02-spatial-data.Rmd
```diff
@@ -652,10 +660,11 @@ extent(new_raster)
 crs(new_raster)
 ```
 
-Unlike the **sf** package, **raster** only use the `proj4string` representation of the coordinate reference system.
+<!--CRSargs(CRS("+init=epsg:4326"))-->
+Note that in contrast to the **sf** package, **raster** only accepts the `proj4string` representation of the coordinate reference system.
 
 Values of the cells are the only non-spatial information in the `Raster*` object.
```
Robinlovelace: I think @Nowosad may be saying that you must spell-out the entire proj4 string, rather than just executing `st_crs(4326)`. I suggest this sentence is rephrased in any case.

02-spatial-data.Rmd
```diff
@@ -774,16 +769,15 @@ r_stack <- stack(raster_in_memory, raster_on_disk)
...
 ```
 
-Due to its properties, `RasterBrick` objects should be processed in a shorter time than `RasterStack`.
-Additionally, operations on `RasterBrick` and `RasterStack` objects will typically return `RasterBrick`.
-On the other hand, `RasterStack` give more flexibility, as a single object could be related to data stored in a memory and on disk in the same time.
-`RasterBrick` objects could be stored only in a memory or on disk.
+Another difference is that the processing time for `RasterBrick` objects should be usually shorter than for `RasterStack` objects.
+Note that operations on `RasterBrick` and `RasterStack` objects will typically return a `RasterBrick`.
 
 ## Coordinate Reference Systems
```
Robinlovelace: Great to have this new raster content - thanks for that @Nowosad and for the comments @jannes-m. My vision is that we make a few basic comments in the next section before a longer section in Chapter 6. Sound good?

02-spatial-data.Rmd
jannes-m: ok, I have already written something on the sticky geometry in chapter 3 (still need to PR this, coming next)

02-spatial-data.Rmd
jannes-m: ok, I have rewritten that to avoid confusion

02-spatial-data.Rmd
jannes-m: Changed that

02-spatial-data.Rmd
jannes-m: Done that.

02-spatial-data.Rmd
jannes-m: Done that.

02-spatial-data.Rmd
Nowosad: My point was that `+init=epsg:4326` is the proj4string representation, while `4326` is just a number (epsg representation). You can have both in sf and only a first one in raster.

jannes-m: actually, I was referring to the raster cell values, but no worries, I've changed that.

02-spatial-data.Rmd
jannes-m: sounds like a plan

02-spatial-data.Rmd
Nowosad: Aha, ok - I get your point. My thinking is that cells have spatial locations, and values of cells don't.

PR #77, opened 2017-08-22
Add roofspace example

Robinlovelace:


Robinlovelace:
The result like this: http://rpubs.com/RobinLovelace/301086

image

PR #79, opened 2017-08-29
Chapter 2

jannes-m:
trying to harmonize chapter 2 and 3 with regard to raster sections

### DIFF
--- a/02-spatial-data.Rmd
+++ b/02-spatial-data.Rmd
@@ -629,9 +629,9 @@ Aside from general raster data manipulation, **raster** provides many low level 
 In this case, **raster** provides the possibility to divide the raster into smaller chunks (rows or blocks), and processes these iteratively instead of loading the whole raster file into RAM (for more information, please refer to `vignette("functions", package = "raster").
 
 For the illustration of the **raster** concepts, we will use datasets from the **spDataLarge** package.
-It consists of a few raster and one vector datasets representing an area of Zion National Park.
+It consists of a few raster and one vector datasets covering an area of the Zion National Park (Utah, USA).
 For example, `srtm.tif` is a digital elevation model of this area (for more details - see its documentation `?srtm`)
-Firstly, we would create a new `RasterLayer` object `new_raster`:
+First of all, we would like to create a `RasterLayer` object named `new_raster`:
 
 ```{r, message=FALSE}
 library(spDataLarge)
@@ -640,7 +640,7 @@ raster_filepath = system.file("raster/srtm.tif", package = "spDataLarge")
 new_raster = raster(raster_filepath)
 ```
 
-If you just type the name of the raster in the console, this will print out the raster header (extent, dimensions, resolution, CRS) and some additional information (class, data source name, summary of the raster values): 
+Typing the name of the raster into the console, will print out the raster header (extent, dimensions, resolution, CRS) and some additional information (class, data source name, summary of the raster values): 
 
 ```{r}
 new_raster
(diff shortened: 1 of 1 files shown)


02-spatial-data.Rmd
```diff
@@ -671,6 +671,7 @@ crs(new_raster)
 <!--CRSargs(CRS("+init=epsg:4326"))-->
 Note that in contrast to the **sf** package, **raster** only accepts the `proj4string` representation of the coordinate reference system.
 
+<!--
```
jannes-m: I present summary functions in chapter 3 (non-spatial attribute data), therefore I have commented this part.

02-spatial-data.Rmd
```diff
@@ -710,66 +712,62 @@ Similar to the **sf** package, **raster** also provides `plot()` methods for its
...
+new_raster2 = raster(nrow = 6, ncol = 6, res = 0.5, 
+                     xmn = -1.5, xmx = 1.5, ymn = -1.5, ymx = 1.5,
+                     vals = 1:36)
 ```
 
-Finally, this function could create `RasterLayer` based on many types of objects, such as a `matrix`, `Extent` or `Spatial*` (from **sp**) object.
-For more information take a look at the help file - `?raster`.
-
+For still further ways of creating a raster object have a look at the help file - `?raster`.
+<!--
```
jannes-m: I have created a raster manually which I will also use in chapter 3. Moreover, I have commented the values, getValues part since I describe these operations in the subsetting and manipulation raster sections. I hope you are ok with it?

02-spatial-data.Rmd
```diff
@@ -913,6 +912,11 @@ The solution is to set the correct units with the **units** package:
 units::set_units(st_area(nigeria), km^2)
 ```
 
+<!-- Is that right? I mean, the units DESCRIPTION says "Support for measurement units in R vectors, matrices and arrays". Since raster datasets are just matrixes, units might be easily used with them?-->
```
jannes-m: @Robinlovelace What do you think?

Robinlovelace: Yes the words below are correct: raster pixel sizes are not given attributes defined by **units**. However, I would replace your line [917](https://github.com/Robinlovelace/geocompr/pull/79/files/dc9edbdb7be85fdb1aa3985da184faf35bbc62bd#diff-237a305a98debc5d406dc19d416f4ce7R917) with the following because **raster** should support **units** not vice versa as currently implied:

```
However, so far **sf** is the only spatial package that supports units, meaning that people working on raster data should approach changes in the units of analysis (e.g. converting pixel widths from imperial to decimal units) with care.
```

Robinlovelace: That also incorporates line 918. I think it would be good to show an example where we calculate the width / height / area of a pixel from a raster dataset already used in this chapter.

02-spatial-data.Rmd
Robinlovelace: I agree with your attempt to minimise duplication. Good work. Will leave it to @Nowosad, who wrote this, to 'sign off' on this change. Fine by me though.

02-spatial-data.Rmd
Nowosad: Ok

02-spatial-data.Rmd
Nowosad: Fine be me for now as well. We could always decide later where it fits better. 
The only question I have for now is about the new raster dataset. Shouldn't we use something we have now? Or (alternatively) add this new data to our data packages?

jannes-m: @Nowosad I created the raster dataset manually, and also described how I did so in the text. I simply replaced the part where you have created a raster object manually with my smaller raster.

02-spatial-data.Rmd
jannes-m: I adapted your wording, and will think of a raster use case tomorrow.

jannes-m: ok, I added a raster unit example

Issue #69, opened 2017-08-07
Add raster attribute operations section to Chapter 3


Robinlovelace:

Robinlovelace:
I think this would fit-in nicely after [section 3.6](http://robinlovelace.net/geocompr/attr.html#removing-spatial-information) and that @jannes-m is well-equipped to tackle this, probably with reference to a simple raster dataset (not sure which one though)...

Nowosad:
https://github.com/Robinlovelace/geocompr/pull/80

PR #80, opened 2017-08-29
Chapter 3

jannes-m:
Work on #69


03-attribute-operations.Rmd
```diff
@@ -29,12 +39,18 @@ The list elements correspond to the number of observations in the attribute tabl
...
+This chapter focuses on non-geographical operations on vector and raster data.
+Regarding vector data, we will introduce subsetting, aggregating or joining attribute data. 
 Note that the corresponding functions also have a geographical equivalent.
 Sometimes you can even use the same functions for attribute and spatial operations.
 This is, for example, the case for subsetting as base R's `[` and tidyverse's `filter()` let you also subset spatial data based on the spatial extent of another spatial object (see Chapter \@ref(spatial-data-operations)).
 That is, the skills you learn here are cross-transferable which is also why this chapter lays the foundation for the next chapter (Chapter \@ref(spatial-data-operations)) which extends the here presented methods to the spatial world.
 
+In the case of raster data we will learn how to create, and how to extract cell values from one layer and multiple layers at once (subsetting).
+Finally, we will introduce the concept of map algebra. 
+Since the underlying mathematics is basically matrix algebra (see above), we present them in this chapter (non-spatial attribute operations).
```
jannes-m: @Robinlovelace @Nowosad 
I think this is justified but might be controversial. What is your opinion on this matter? Of course, one could also argue to present map algebra in chapter 4. Or we could also think about rearranging the chapter structure. For instance, instead of chapter 3 Non-spatial attributes and 4 Spatial operations, we could have chapter 3 on vector data including non-spatial and spatial operations and chapter 4 on just raster data.

03-attribute-operations.Rmd
```diff
@@ -499,14 +507,211 @@ world_data = world %>% st_set_geometry(NULL)
...
+Also, terrain processing uses heavily focal functions.
+Think, for instance, of the calculation of the slope, aspect and flow directions.
+The `terrain()` function lets you compute a few of these terrain characteristics but has not implemented all popular methods
+For example, the Zevenbergen and Thorne method to compute the slope is missing.
+Equally, many other terrain and GIS functions are **not** implemented in R such as curvatures, contributing areas, wetness indexes, and many more.
+Fortunately, desktop GIS commonly provide these algorithms.
+In Chapter 13 we will learn how to access GIS functionality from within R.
+<!-- Reference 13-gis chapter -->
+
+**Zonal and global operations**
```
jannes-m: Not finished the rest yet, still have to write this. Will do so tomorrow.

code/03_focal_example.R
```diff
@@ -0,0 +1,90 @@
...
+poly_target =
+  rbind(c(-1, 0.5), c(-0.5, 0.5), c(-0.5, 1), c(-1, 1), c(-1, 0.5)) %>%
+  list %>%
+  st_polygon %>%
+  st_sfc %>%
+  st_sf(data.frame(id = 1), geometry = ., crs = 4326) %>%
+  as(., "Spatial")
+
+#**********************************************************
+# 2 FOCAL EXAMPLE------------------------------------------
```
jannes-m: @Nowosad Yes, I have also looked at **visualraster** but was not entirely convinced. But I admit, I am a bit biased towards **lattice**. But I guess it doesn't matter which code we use here, lattice or ggplot2, since we don't show it in the book, and the results of both packages should be identical.

03-attribute-operations.Rmd
Nowosad: That a tough decision to make. Extracting values is definitely "non-spatial" thing, I would assume that zonal operations are "spatial", and map algebra is somewhat in the grey zone... 

@Robinlovelace, what do you think about this?

jannes-m: Somewhere above in the text I explain:
`The raster header simply gives the matrix a spatial dimension which we need when plotting the raster or when we want to combine two rasters, think, for instance, of adding the values of one raster to another.
Before such an operation, the corresponding algorithm simply checks if the two rasters share the same extent, projection and resolution, that is the raster headers. 
Hence, the spatial dimension of raster data is (often) of no importance at the processing level.
Then, raster processing uses in fact matrix algebra which is why it is so much faster compared to vector processing.
`
It doesn't matter if is local, focal, zonal or global (all of these operations are part of the map algebra definition), it is just matrix algebra. Zones are not identified via coordinates but via indices, say `mat[c(1:3, 4:7, 12:18), c(24:28, 77:100)]` or cell IDs - nothing spatial about this.
But yes, you are right, might be controversial. On the other hand, when we can explain why we putting it here, you can't really argue with it.
Anyways, it's a tough call as you say, and that's why I also proposed to simply separate vector and raster into two chapters.

Robinlovelace: Interesting discussion and sorry for arriving late to it. I agree that matrix algebra is important when thinking about rasters *in depth* but it's not essential to know that that's why raster datasets are conducive to fast, scalable and parallel processing. I agree that a raster is a very different beast compared with a vector dataset however, having given it some thought, I suggest we don't split-out the raster stuff into a new chapter at this stage.

I think the amount of effort to restructure and re-word everything around this new way of doing things would take-up a fair amount of time, while there are clearly 'lower hanging fruit' in terms of finishing a cohesive draft ready for the reviewers (Chapter 4 is in a sorrier state than Chapter 3 atm, and Chapter 5 could definitely benefit from some work). I think it's a rather 'GIS-y' structure to cover the vector and raster data separately but I think it's good to consider them all as spatial datasets which you can do.

It does make me think about the importance of raster-vector conversion and operations - (see https://stat.ethz.ch/pipermail/r-sig-geo/2017-August/025937.html ) - and suggest we put in a section on Raster-vector operations in Chapter 4 - would fit nicely in there now I've re-ordered things there is plenty of space (I would say 9 sections per chapter is enough).

Robinlovelace: Plus have just added a question to reviewers on this: https://github.com/Robinlovelace/geocompr/commit/de26707cdf1efe927e6da0a8b826842f1bb489b6

03-attribute-operations.Rmd
Robinlovelace: Great piece of continuous technical prose.

03-attribute-operations.Rmd
```diff
@@ -499,14 +507,211 @@ world_data = world %>% st_set_geometry(NULL)
...
+To make a spatial prediction, all we have to do, is to apply the estimated coefficients to the predictor rasters, and summing up the resulting output rasters (Chapter ??; see also @muenchow_predictive_2013).
+<!-- add reference to chapter ecological modeling -->
+
+While local functions operate on one cell, though possibly from multiple layers, **focal** operations take into account a central cell and its neigbors.
+The neigborhood (also named kernel, filter or moving window) under consideration is typically of size 3-by-3 cells (that is the central cell and its eight surrounding neighbors) but can take on any other (not necessarily rectangular) shape as defined by the user.
+A focal operation applies an aggregation function to all cells within the specified neighborhood, uses the corresponding output as the new value for the the central cell, and moves on to the next central cell (Figure @ref(fig:focal_example)).
+Other names for this operation are spatial filtering and convolution [@burrough_principles_2015].
+
+In R, we can use the `focal()` function to perform spatial filtering.
```
Robinlovelace: Good description of focal operations (are they spatial or attribute thought ; ) I think they are fine here but could also fit in Chapter 4. Let me know when your teeth into that chapter if you think it fits...

03-attribute-operations.Rmd
```diff
@@ -499,14 +507,211 @@ world_data = world %>% st_set_geometry(NULL)
...
+We define the shape of the moving window with a `matrix` whose values correspond to weights.
+Secondly, the the `fun` argument lets us specify the function we wish to apply to this neighborhood.
+Here, we choose the minimum, but of course we can use any other function such as the the sum, the mean, the median, the mode, the maximum or the variance.
+
+```{r, eval = FALSE}
+r_focal = focal(r, w = matrix(1, nrow = 3, ncol = 3), fun = min)
+```
+
+```{r focal_example, echo = FALSE, fig.cap = "Input raster (left) and resulting output raster (right) due to a focal operation - summing up 3-by-3 windows"}
+knitr::include_graphics("figures/03_focal_example.png")
```
Robinlovelace: This is an excellent figure, great work.

Robinlovelace:
Just looking over this PR and thinking it's looking great. It would be great to have this ready for the workshop on Monday so would like to merge this Friday morning. Great to see you getting stuck-in with writing too - there will be much more of that after these first 5 chapters are off to reviewers and that will be really fun.

03-attribute-operations.Rmd
jannes-m: Yes, we could add them to Chapter 4 as well. In any case, I would recommend not to spread the different pillars of map algebra into different chapters, i.e. if we put the focal stuff into c4 we should do so with the local, zonal and global operations. I guess, this is also what most readers would expect. 
Thanks for adding the raster question to the list for the reviewers. Good idea to move on in the meantime 😃

03-attribute-operations.Rmd
jannes-m: Thanks, glad you like it!

jannes-m:
Ok, I have also written the paragraphs on zonal and global operations. I also added a few raster exercises. I think this should be enough here? Or do you have any further recommendations? I think I will refine tiny bits tomorrow but I can always issue an new PR if you are in a hurry.

PR #81, opened 2017-09-03
Chapter 3/4

jannes-m:
Hey there,

@Nowosad, alter Haudegen (German :)), I have managed to restructure and rewrite c3/c4 a bit after the wedding (what was a bit hard with watering eyes....).  However, it is not perfect, but I guess good to go for the workshop, after all the book is work in progress.
I guess, Robin, already moved the map algebra stuff to c4.


Nowosad:
I'm merging it now and going to check if everything is working in the evening. Life's too short to wait.

Robinlovelace:
Fantastic stuff @jannes-m thanks for tidying-up after my shift of the map algebra to c4 - all the re-structuring we're doing now will pay off now - still up for any other ideas in that direction. I'll go over c3 and create a print-our for the workshop - should do the (2nd I think) release of this repo also as watershed, just before this gets sent for review.

PR #84, opened 2017-09-07
Chapter 4

jannes-m:
@Robinlovelace I have not read to the end of the vector stuff but I still push this now so that you possibly can take into account my suggestions while working on c4. The main things so far:

1. I have deleted a few sentences, because they simply repeated stuff from c3, I hope that's ok.
2. I wouldn't separate spatial subset operations and all the other topological operations. We don't have to show them in detail. The topological figure is really great, and should explain what the difference is between the several operations.
3. Be careful with the aggregating/dissolving stuff! aggregate and dplyr::summarize are doing a spatial dissolving operation (look at the comments below).


04-spatial-operations.Rmd
```diff
@@ -22,17 +23,21 @@ Spatial operations are an important component of any geospatial software and vit
...
 In the case of vector data, each of these spatial operations has a non-spatial equivalent, as demonstrated in section \@ref(vector-attribute-manipulation) in the previous chapter.
-Of course, spatial raster and vector operations share many similarities, at least in terms of functionality.
-The internal processing is completely different. 
-In contrast to vector operations, subsetting is the only spatial raster operation we present in this chapter (section \@ref(spatial-operations-on-raster-data)) which also has a non-spatial counterpart.
+By contrast, subsetting is the only spatial raster operation we present in this chapter (section \@ref(spatial-operations-on-raster-data)) which also has a non-spatial counterpart.
+Nevertheless, spatial raster and vector operations share many similarities, at least in terms of functionality though the internal processing is completely different.
+Be careful with the wording.
+Sometimes the same words have slightly different meanings for both data models.
+<!-- maybe we should choose dissolving as the term here, aggregate is more R lingo-->
```
jannes-m: Well, one to think about. I guess, aggregate is more R lingo, in the GIS world you talk of dissolving. Or we just have to make clear that aggregate and dissolve means the same when it comes to vector data.

04-spatial-operations.Rmd
```diff
@@ -41,53 +46,52 @@ It is important to note that spatial operations that use two spatial objects rel
 ## Spatial subsetting
 
 Spatial subsetting is the process of selecting only those features of a spatial object that in some way *intersect* with another spatial object.
+<!-- can we already reference the intersection figure here? Generally asking, why can't we present the other topological operations here? Would fit nicely in this section... -->
```
jannes-m: Yes, I wouldn't separate spatial subsetting and the other topological operations. They just fit fine here. Moreover, we need also need them for spatial joins so we should present them here already! Maybe we have then to rename the header of the subsection.

04-spatial-operations.Rmd
```diff
@@ -202,13 +205,15 @@ plot(joined["Population (millions)"], add = TRUE)
 par(mfrow = c(1, 1))
 ```
 
-Like attribute data aggregation, covered in section \@ref(vector-attribute-aggregation), spatial data aggregation can be a way of *condensing* data.
+<!-- I would expand it a little bit on spatial joins. Typically, this operation is also known as spatial overlay. Of course, you can also use other geometries (lines), and especially you can not only use st_intersects (another reason why the topological stuff should have been already covered before). Also interesting would be to show what happens if there are multiple matches, e.g., multiple points per polygon. In sp, this was solved using a list containing the IDs of the matched objects while not returning the corresponding attribute table. Overall, a comparison with sp::over would be also interesting -->
```
jannes-m: Yes, I would expand on spatial joins, see note for more explanation!

04-spatial-operations.Rmd
```diff
@@ -202,13 +205,15 @@ plot(joined["Population (millions)"], add = TRUE)
...
 Aggregated data show some statistic about a variable (typically mean average or total) in relation to some kind of *grouping variable*. 
 For attribute data aggregation the grouping variable is another variable, typically one with few unique values relative to the number of rows.
 The `continent` variable in the `world` dataset is a good example:
-there are only 8 unique continents but 177 countries.
-In section \@ref(vector-attribute-aggregation) the aggregation process condensed the `world` dataset down into only 8 rows and an aggregated `pop` variable representing the total population per continent (see Figure \@ref(fig:continent-pop)).
-
+there are only eight unique continents but 177 countries.
+In section \@ref(vector-attribute-aggregation) the aggregation process condensed the `world` dataset down into only eight rows and an aggregated `pop` variable representing the total population per continent (see Figure \@ref(fig:continent-pop)).
+<!-- CAREFUL, in fact, fig:continent-pop has already been a spatial operation (dissolving polygons). You should have shown the table only in the previous section -->
```
jannes-m: Be careful! I guess fig:continent-pop belongs here because it was a spatial operation! We should have just shown the table in c3!

04-spatial-operations.Rmd
```diff
@@ -217,8 +222,16 @@ Since mid-2017 (with the release of **sf** `0.5-3`) the base R function `aggrega
 Building on the example presented the previous section (\@ref(spatial-subsetting)), we demonstrate this by aggregating the population of countries that intersect with the buffer represented by the circular `buff` object created in the previous section.
 
 ```{r}
-buff_agg = aggregate(x = africa["pop"], by = buff, FUN = sum)
+buff_agg = aggregate(x = africa[, "pop"], by = buff, FUN = sum)
+```
+<!--
+show also tidyverse way, so what you are doing is basically a spatial join and a subsequent aggregation without a grouping variable. Didactically, it might be better to present a grouping variable.
```
jannes-m: Well, I am sorry to criticize your nice example! But aggregating is just a synonym for dissolving, and this is more impressive (especially spatially) if we have a grouping variable.

04-spatial-operations.Rmd
```diff
@@ -217,8 +222,16 @@ Since mid-2017 (with the release of **sf** `0.5-3`) the base R function `aggrega
...
+```{r}
+st_join(buff, africa[, "pop"]) %>%
```
jannes-m: note that this is the tidyverse version of the aggregate command. So basically, you are doing a spatial join and then summarize all data without a grouping variable...

04-spatial-operations.Rmd
```diff
@@ -2,15 +2,16 @@
 
 ## Prerequisites {-}
 
-- This chapter requires the packages **tidyverse**, **sf** **raster**, and **spData**:
+- This chapter requires the packages **tidyverse**, **sf** and **raster**. 
+To avoid confusing function names (e.g., `dplyr::select` and `raster::select`) we will wait with attaching the **raster** package until we reach section \@ref{spatial-operations-on-raster-data):
```
Robinlovelace: Reword to fix English grammar. I suggest:

```
To avoid confusing function names (e.g., `dplyr::select` and `raster::select`) we will wait until section \@ref{spatial-operations-on-raster-data) between loading **raster**:
```

Also note: we have not explained what `::` is yet - in my comment I explain it before referring to functions in this way. Maybe a good candidate for a note? Also I think it's good to *always* use the `dplyr::select()` because it will likely (and frequently does for me) cause problems later down the line. So in summary I'm not in favour of this change.

04-spatial-operations.Rmd
```diff
@@ -2,15 +2,16 @@
...
 
 ```{r, message=FALSE}
 library(sf)
-library(raster)
 library(tidyverse)
+# library(raster)
 ```
 
-- Please attach also the **spData** package which automatically attaches the `world` dataset:
+- Please attach also the **spData** package which in turn automatically attaches the `world` dataset:
```
Robinlovelace: I don't think the word 'please' is appropriate here, overpolite! (May have been my English politeness that introduced it but can't remember.)

04-spatial-operations.Rmd
```diff
@@ -22,17 +23,21 @@ Spatial operations are an important component of any geospatial software and vit
...
 In the case of vector data, each of these spatial operations has a non-spatial equivalent, as demonstrated in section \@ref(vector-attribute-manipulation) in the previous chapter.
-Of course, spatial raster and vector operations share many similarities, at least in terms of functionality.
-The internal processing is completely different. 
-In contrast to vector operations, subsetting is the only spatial raster operation we present in this chapter (section \@ref(spatial-operations-on-raster-data)) which also has a non-spatial counterpart.
+By contrast, subsetting is the only spatial raster operation we present in this chapter (section \@ref(spatial-operations-on-raster-data)) which also has a non-spatial counterpart.
+Nevertheless, spatial raster and vector operations share many similarities, at least in terms of functionality though the internal processing is completely different.
+Be careful with the wording.
```
Robinlovelace: This should be commented before the PR is merged and the changes go live!

04-spatial-operations.Rmd
Robinlovelace: I think dissolve is the spatial operation that accompanies the attribute aggregation. Definitely an improvement the way you've worded it - may reword it slightly.

04-spatial-operations.Rmd
Robinlovelace: From a teaching perspective it's much easier to learn things with one operation first and then discover there are many you can use. A cross-reference here would be ideal.

04-spatial-operations.Rmd
```diff
@@ -41,53 +46,52 @@ It is important to note that spatial operations that use two spatial objects rel
...
 Spatial subsetting is the process of selecting only those features of a spatial object that in some way *intersect* with another spatial object.
+<!-- can we already reference the intersection figure here? Generally asking, why can't we present the other topological operations here? Would fit nicely in this section... -->
 Note that 'intersect' in this context has a precise meaning:
 if `y` is used to subset features in a 'target' object of `x`, any features in `x` that touch, overlap or are within features in `y` will be selected.
 Intersect is the default operation for spatial subsetting but others can be used using the `op =`
 argument.^[Interested
 readers can see this default value of `op` set in the first line of the function call by entering its long-form name into the console: 
-`` sf:::`[.sf` ``
+`` sf:::`[.sf` `` 
+The `?sf` help page documents this also.
```
Robinlovelace: Does this footnote work OK? if so, great addition but I think there should be a period before the capital T in The.

04-spatial-operations.Rmd
```diff
@@ -41,53 +46,52 @@ It is important to note that spatial operations that use two spatial objects rel
...
 after you understand how to spatially subset objects that *intersect* another (via `st_intersects()`) it is easy to subset based on other types of spatial operation such as `st_touches()`, `st_crosses()` and `st_within()`.
 For this reason now we focus only on one of the spatial subsetting operations.
 We use `st_intersects()` instead of any of the others not only because it the default when subsetting with `[`,
 but also `st_intersects()` is useful as a 'catch all' that identifies all types of spatial relations.
 
 In general terms, spatial subsetting is simply the spatial equivalent of *attribute subsetting*.
-However, to do spatial subsetting *two spatial objects are needed* the spatial relation between which is to be established.
 As with attribute subsetting, spatial subsetting is a *binary operation*: an object is either selected or not.
-As in section \@ref(vector-attribute-subsetting), we start with base methods before describing how to do it in the **tidyverse**.
+Following the structure of section \@ref(vector-attribute-subsetting), we start with base methods before describing how to do it in the **tidyverse**.
```
Robinlovelace: Definite improvement, thanks for that.

04-spatial-operations.Rmd
```diff
@@ -41,53 +46,52 @@ It is important to note that spatial operations that use two spatial objects rel
...
+To introduce spatial vector subsetting operations we will use countries in Africa.
+We will apply attribute subsetting to the `world` dataset (see previous chapter):^[Recall
 that we can also subset simple features using the `filter()` function, e.g. with `filter(world, continent == "Africa")`]
 
 ```{r}
 africa_wgs = world[world$continent == "Africa", ]
 ```
 
-To further set-up the input data, we will reproject the data to the coordinate reference system (CRS) 32630 (it's EPSG code, explained in Chapter 6):
+To further set up the input data, we will reproject the data to the coordinate reference system (CRS) 32630, its EPSG code (explained in Chapter 6):
```
Robinlovelace: I think setup is better - we should be consistent so worth searching for all instances. Set up sounds strange and stilted to me. They use setup in R4DS so let's go with that: http://r4ds.had.co.nz/introduction.html

04-spatial-operations.Rmd
```diff
@@ -41,53 +46,52 @@ It is important to note that spatial operations that use two spatial objects rel
...
+We can also use the `[` operator for *Spatial* subsetting.
+The difference is that we use *another spatial object* inside the square brackets instead of an `integer` or `logical` vector.
 This is a concise and consistent syntax, as shown in the next code chunk.
 Let's test it with a hypothetical scenario: we want to subset all countries within 20 degrees of the point where the equator (where latitude = 0 degrees) intersects the prime meridian (longitude = 0 degrees), as illustrated in Figure \@ref(fig:globe).
 The subsetting object is created below.
 Note that this must have the same CRS as the target object (set with the `crs` argument):
 
 ```{r, warning=FALSE}
-center = st_sf(st_sfc(st_point(c(0, 0)), crs = 4326))
+center = st_sf(geometry = st_sfc(st_point(c(0, 0)), crs = 4326))
```
Robinlovelace: Yes, important to be explicit. Thanks for that.

04-spatial-operations.Rmd
```diff
@@ -165,15 +168,15 @@ For further discussion of this decision, and some controversy, see the (closed)
 ```
 
 ```{r}
-row.names(africa[africa$subregion == "Northern Europe", ])
-row.names(filter(africa, subregion == "Northern Europe"))
+row.names(africa[africa$subregion == "Northern Africa", ])
```
Robinlovelace: My bad. Thanks for the fix.

04-spatial-operations.Rmd
Robinlovelace: OK - could you do that?

jannes-m: Sure, there's only one problem. Maybe you remember, we have discussed this before: Dissolving the world led to some strange artifacts in the output (borders within continents where no borders should be). I have checked this is due to an unclean topology, here overlapping polygons. I have tried to clean this using GRASS. This makes it better but not perfect. So the question is if we cannot use a clean world map in the first place? But maybe that's not that easy, I have tested also other world maps available in R and these were also unclean... We can also use just another dataset, e.g.,

```r
library(spData)
group_by(us_states, REGION) %>% 
   summarize(pop = sum(total_pop_15)) %>% 
   select(pop) %>%
   plot
```

04-spatial-operations.Rmd
jannes-m: I agree. But I meant to just append the topological stuff to the spatial subsetting and not separate it into two subsections with other subsections in between.

04-spatial-operations.Rmd
jannes-m: You have already written a note on this in c3. Can change this back but then we will probably also have to always use `dplyr::filter()`

Robinlovelace: I've never had name clashes with filter - don't think it's a raster function...

Robinlovelace: Agree it's kinda inconsistent to only do it with select though. Will ask @Nowosad who's right here!

04-spatial-operations.Rmd
Robinlovelace: I see your point:

``` r
library(spData)
library(sf)
#> Linking to GEOS 3.5.1, GDAL 2.2.1, proj.4 4.9.2, lwgeom 2.3.3 r15473
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
group_by(world, continent) %>% summarize(pop = sum(pop)) %>% select(pop) %>% 
  plot
```

Robinlovelace: Using the US data instead sounds fine but we should mention how to fix it.

Robinlovelace: But we should also so how to fix it, as illustrated below (thanks @Nowosad - pls link to issue):

``` r
library(spData)
library(sf)
#> Linking to GEOS 3.5.1, GDAL 2.2.1, proj.4 4.9.2, lwgeom 2.3.3 r15473
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
world %>% st_set_precision(precision = -1) %>% group_by(continent) %>% summarize(pop = sum(pop)) %>% 
  select(pop) %>% plot
```

jannes-m: Well, that's only a quick-and-dirty solution since this doesn't clean the topology, and overlapping polygons still remain 😄 . But yes, good point, we should mention, that setting the precision sometimes facilitates geometrical operations with GEOS.

Nowosad: @jannes-m The question is if there is a good way to clean a topology in R? Maybe it could be a nice small section in the book.

04-spatial-operations.Rmd
Robinlovelace: I think aggregating is different from dissolving. The former operates on attributes the latter on the geometry. They can work well together, good to clarify the distinction.

04-spatial-operations.Rmd
Nowosad: There isn't a right answer here in my opinion. We should decide on a one way of doing that though.

04-spatial-operations.Rmd
Robinlovelace: BTW yes I'm well-up for another example. Any ideas? The thinking behind this one was that it was clear and intuitive. It does not do dissolving. Maybe we could a second aggregating example with multiple features in the 'source' object.

04-spatial-operations.Rmd
Robinlovelace: That's one way of putting it. Spatial aggregation is a type of spatial join.

04-spatial-operations.Rmd
Robinlovelace: Another note: without `dplyr::` it was failing locally, even when I loaded **raster** later. So my preference is to keep it that way - minimises confusion all round I think.

04-spatial-operations.Rmd
jannes-m: Well, I would say you selected a 'special' example. The `aggregate` function actually dissolves polygons. However, your input `buff` was only one polygon, so there was nothing to dissolve. 
Check this out:
```r
plot(aggregate(africa[, "pop"], by = list(africa$subregion), FUN = sum, na.rm = TRUE))
# which is the same as:
plot(group_by(africa, subregion) %>% summarize(pop = sum(pop, na.rm = TRUE)))
```
See, what I mean. 
And yes, you could also give a spatial object to the by command, and what you get back, is the aggregate of the intersection. So this might be a bit confusing, which is I would skip this, and stick with a real dissolving example (like here the africa example or the us_states examples). We can show first how to do it in base R (`aggregate`) and then by using `summarize()`

04-spatial-operations.Rmd
jannes-m: the point here is, if you don't name the column `geometry`, it will be named automatically in a weird way.

04-spatial-operations.Rmd
jannes-m: Ok, perfect.

04-spatial-operations.Rmd
jannes-m: Depends on the input, as usual, see also Barry's answer to this [post](http://r-sig-geo.2731867.n2.nabble.com/Cleaning-up-self-intersections-td7581945.html). 
My usual approach is to use rgrass7. `writeVECT`  uses `v.in.ogr` and this cleans the vector geometry and builds a topology. And then I simply load the cleaned data back into R using `readVECT`. In many cases this works, however, if the input data is too unclean, then GRASS can't solve this either easily. This was the case with the `world` dataset.

04-spatial-operations.Rmd
Robinlovelace: Perhaps the example should be Chapter 3 as the operation does not change the geometry of either object. I think I know what you mean... The example is deliberately simple, to show how the function `aggregate` works. I think a second example should build on this one on the basis that you should start simple and add complexity. Suggestions along those lines welcome e.g. via updates to this PR.

jannes-m: It does change the geometry, see the polygon boundaries (just 5 polygons left), and compare them to `plot(africa[, "pop"])` (51 polygons). In fact, that's the whole point of dissolving. Somehow, I have the impression we talk at cross-purposes... sorry, if that's the case!

jannes-m: Just for comparison purposes:
```r
par(mfrow = c(1, 2))
plot(africa[, "pop"])
group_by(africa, subregion) %>% 
  summarize(pop = sum(pop, na.rm = TRUE)) %>%
  select(pop) %>% 
  plot
```

Robinlovelace: Yes we were talking about different things: I was talking about the example I added : )

Your example clearly changes the geometry. I was referring to the buffer example when I said it could go in Chapter 3. What do you think?

jannes-m: Ah, ok, then this is resolved 😄 . 
If we are talking about `aggregate(x = africa[, "pop"], by = buff, FUN = sum)`, I wouldn't put it into c3, since it is doing a spatial join (using st_intersect in the background), so I would put it to the spatial join subsection. I will be off now helping a friend with painting his new house! So have fun and read and hear you soon!
BTW, I have also some loose thoughts how to structure the book, will put them onto paper within the next few days, and give this to you as a basis for discussion.

04-spatial-operations.Rmd
```diff
@@ -77,7 +77,7 @@ that we can also subset simple features using the `filter()` function, e.g. with
 africa_wgs = world[world$continent == "Africa", ]
 ```
 
-To further set up the input data, we will reproject the data to the coordinate reference system (CRS) 32630, its EPSG code (explained in Chapter 6):
+To further setup the input data, we will reproject the data to the coordinate reference system (CRS) 32630, its EPSG code (explained in Chapter 6):
```
Robinlovelace: Ahh in this context the single word version does sound a bit strange to me. I think `setup` is a noun. Therefore `set up` would be fine by I suggest `prepare` would be better still. Cheers for updating the PR!

jannes-m: Done that.

04-spatial-operations.Rmd
jannes-m: changed that

04-spatial-operations.Rmd
jannes-m: changed that

04-spatial-operations.Rmd
```diff
@@ -277,11 +276,13 @@ These issues can be tackled when aggregating areal target data with areal interp
 
 Spatial congruence is an important concept related to spatial aggregation.
 An *aggregating object* object (which we will refer to as `y`, representing the buffer object in the previous section) is *congruent* with the target object (`x`, representing the countries in the previous section) if the two objects have shared borders.
-Often this is the case for administrative boundary data, whereby the larger units (e.g. Middle Layer Super Output Areas in the UK) are composed of many smaller units (Output Areas in this case, see [ons.gov.uk](https://www.ons.gov.uk/methodology/geography/ukgeographies/censusgeography) for further details).
+<!-- this really is very special in the UK, most other European countries have districts and municipalities but there is also the official NUTS classification for EU member states -:); maybe this is a more general classification scheme, see http://ec.europa.eu/eurostat/web/nuts/overview. I learned that when I was at geomarketing.-->
```
jannes-m: maybe, we should use a more general example such as the NUTS regions (see comment)

04-spatial-operations.Rmd
```diff
@@ -277,11 +276,13 @@ These issues can be tackled when aggregating areal target data with areal interp
...
+Often this is the case for administrative boundary data, whereby the larger units (e.g., Middle Layer Super Output Areas in the UK) are composed of many smaller units (Output Areas in this case, see [ons.gov.uk](https://www.ons.gov.uk/methodology/geography/ukgeographies/censusgeography) for further details).
 
 *Incongruent* aggregating objects, by contrast, do not share common borders with the target [@qiu_development_2012].
 This is problematic for spatial aggregation (and other spatial operations) illustrated in Figure \@ref(fig:areal-example).
 Areal interpolation resolves this issue.
+<!-- well, pycnophylactic interpolation has been developed for spatial downscaling, e.g., predicting/interpoling the number of inhabitants at a finer spatial level (municipality) on the basis of inhabitants of a coarser spatial level (districts).-->
```
jannes-m: Be careful when it comes to spatial downscaling and/or spatial aggregation.

04-spatial-operations.Rmd
```diff
@@ -303,7 +304,7 @@ detach("package:tmap", unload = TRUE)
 
 The simplest useful method for spatial interpolation is *area weighted* spatial interpolation.
 This is implemented in `st_interpolate_aw()`, as demonstrated below:
-
+<!-- well, what you are doing here is weighting the inhabitants presumably on the basis of the area. The link between your congruent and incongruent example is somehow missing. One gets the impression, that the spatial interpolation takes care of incongruencies which is not the case. So I would suggest to clarify the weighting procedure.-->
```
jannes-m: I would suggest to make clear that `st_interpolate_aw()` not sums up the inhabitants but weighs them by area. And yes, this is a simple approach and better than just to sum up the inhabitants but imagine your circle just intersects the desert parts of a country.

04-spatial-operations.Rmd
Robinlovelace: Yes great plan, especially in the context of Brexit!

Robinlovelace: Happy for you do do something on that - you up for adding it (can be after this PR is merged...).

PR #85, opened 2017-09-13
Update chapter plan

Robinlovelace:
Feedback welcome @jannes-m and @Nowosad

### DIFF
--- a/our_chapters.md
+++ b/our_chapters.md
@@ -9,14 +9,13 @@ Foundations
 
 Applications
 
+- introductory transport rl
 - ecological - jm
 - location-analysis - jm
-- transport rl
+- land-use analysis - jn
 
 Advanced methods
 
-- advanced raster
-- advanced vector 
 - raster-vector - jn
 - spatial-crossvalidation - jm
 - advanced-visualization - jn


jannes-m:
Thank you for your thoughts and input! Regarding the chapter structure we have slightly different opinions but of course we will agree on something eventually. After all, discussions make us think more thoroughly, and yes I admit I feel unhappy if the structure is not round (not sure, if you can say so in English). But yes, it is work in progress and it will develop naturally eventually. Additionally, I only know what I would like to write in the use cases but have only a slight idea of your uses cases. Still I think it helps to discuss. And believe me I understand the advantage of your approach. Nevertheless, I have a few questions:

1. What do we do if we need the advanced methods already in the application chapters (for instance spatial cross validation for ecological modeling and landcover classification; I will also need R-GIS bridges for the ecological modeling)? Should we then point to the advanced chapters?
2. What will be the contents of the advanced raster (the landcover classification would be a nice example for this) and vector chapters?
3. Where do we fit in big data/cloud computing? Or don't you want to present this?

What do you think about this compromise? Let's lay the foundations followed by a few use cases (2-3). Subsequently, we present the advanced methods which again are followed by 2-3 use cases. This would also have the advantage that R-GIS bridges are presented much earlier which I would prefer. After all, we are using R as a GIS all the time and a natural next step would be to access a GIS from within R.

Robinlovelace:
Hi Jannes, it seems that having 4 'parts' with basic examples and then more advanced examples would overcome your 3 issues:

1. Yes it will be great to use the R-GIS bridges stuff in an applied setting and agree the applied section would have to go after - in 'part 4' if we split it into 4 parts.
2. Note that this PR [removes ](https://github.com/Robinlovelace/geocompr/pull/85/files#diff-f136a9ee7e3c7ad327980e034085b361L18)'Advanced Raster' and 'Advanced Vector' as chapters.
3. I think the big data chapter could be one we keep 'on the backburner' - we can decide whether or not to go for it after. Also one to ask the reviewers. I think it's a good idea.

I plan to update the plan in this branch to reflect these points so we can continue discussion which as you say is vital to getting something that is cohesive and well-structured.

Robinlovelace:
Here's a beginning of an alternative:

Introduction

- introduction - rl

Foundations

- spatial-data - rl
- attr - jn
- spatial - jn
- read-write - rl
- transform - jn

Applications

- introductory transport rl
- location-analysis - jm
- land-use analysis - jn

Advanced methods

- raster-vector - jn
- spatial-crossvalidation - jm
- advanced-visualization - jn
- algorithms-functions - rl
- gis - jm

Advanced applications

- ecological - jm
- renewable energy (?)
- sea level rise (?)

Conclusion and next steps

PR #86, opened 2017-09-14
c3 and 4 again

jannes-m:
I have restructured the content and also modified the aggregation and dissolving example. However, you would still need to update the aggregate circle example, I wasn't really sure how to fit it in elegantly.


03-attribute-operations.Rmd
```diff
@@ -232,25 +232,31 @@ Another advantage over the nesting approach is that you can easily comment out c
...
+Note that you have to indicate the grouping variable as a `list`-object:
 
-<!-- https://github.com/ropenscilabs/skimr ?? -->
-
-<!-- As demonstrated in chapter \@ref(spatial-class), `summary()` provides a quick summary of the spatial and non-spatial components of spatial objects.
-Enter the following command for an overview of the `world` object and all its variables (result not shown):
-
-```{r, eval=FALSE}
-summary(world)
+```{r, eval = FALSE}
```
jannes-m: I have swapped places, i.e. I start with a base R example followed by a tidyverse example, just to remain consistent.

04-spatial-operations.Rmd
```diff
@@ -176,13 +170,133 @@ row.names(africa[africa$subregion == "Northern Africa", ])
 row.names(filter(africa, subregion == "Northern Africa"))
 ```
```
jannes-m: I have placed the topological relations directly after the subsetting operations because subsetting is actually a subgroup of the topological operations. Nevertheless, I would suggest to just show the intersection example, and maybe one further example. I think referencing the figure showing the differences between the topological operations should suffice.

04-spatial-operations.Rmd
```diff
@@ -209,21 +324,60 @@ plot(joined["population_millions"], add = TRUE)
...
+The `REGION` variable in the `us_states` dataset is a good example:
+there are only four subregions but 49 states (excluding Hawaii and Alaska).
+In section \@ref(vector-attribute-aggregation) we have already seen how attribute aggregation process condensed the `world` dataset down into only eight rows.
+Spatial data aggregation is the same conceptually but in addition to aggregating the attribute data, it dissolves the underlying polygons.
+Here, we want to aggregate the state population into regions.
+This means that we not only end up with four rows but also with four polygons (out of 49 in the beginning).
+As with spatial subsetting, spatial aggregation operations work by extending existing functions:
+
+```{r}
+regions = aggregate(x = us_states[, "total_pop_15"], by = list(us_states$REGION),
```
jannes-m: I have tried the africa example but there again were those ugly artefacts we have already observed when doing the aggregation on the world dataset. Therefore, I chose the us_states.

04-spatial-operations.Rmd
```diff
@@ -209,21 +324,60 @@ plot(joined["population_millions"], add = TRUE)
...
+
+Of course, there is also spatial tidyverse counterpart.
+You can achieve the same with:
+
+```{r, eval = FALSE}
+group_by(us_states, REGION) %>%
+  summarize(sum(pop = total_pop_15, na.rm = TRUE))
+```
+
+<!--Not sure how to elegantly include your circle buffer intersection example -->
```
jannes-m: Yep, I was not really sure how to elegantly include the following example. But I guess, if you shortened it substantially, it could be a good transition towards your congruence subsection. However, one has to be a bit careful not to confuse the reader so one has to make clear that this example shows an aggregation without dissolving anything. Maybe a nice idea, because we did so in c3 with the world dataset as well.

04-spatial-operations.Rmd
```diff
@@ -279,13 +433,12 @@ These issues can be tackled when aggregating areal target data with areal interp
...
-<!-- this really is very special in the UK, most other European countries have districts and municipalities but there is also the official NUTS classification for EU member states -:); maybe this is a more general classification scheme, see http://ec.europa.eu/eurostat/web/nuts/overview. I learned that when I was at geomarketing.-->
-Often this is the case for administrative boundary data, whereby the larger units (e.g., Middle Layer Super Output Areas in the UK) are composed of many smaller units (Output Areas in this case, see [ons.gov.uk](https://www.ons.gov.uk/methodology/geography/ukgeographies/censusgeography) for further details).
+Often this is the case for administrative boundary data, whereby the larger units (e.g., Middle Layer Super Output Areas in the UK or districts in many other European countries) are composed of many smaller units (Output Areas in this case, see [ons.gov.uk](https://www.ons.gov.uk/methodology/geography/ukgeographies/censusgeography) for further details or municipalities in many other European countries).
 
 *Incongruent* aggregating objects, by contrast, do not share common borders with the target [@qiu_development_2012].
```
jannes-m: I was not sure if we really should open this barrel here ( -:), Germlish) because you could write a book about interpolation. Maybe, you could say that it was more just to use an areal weighted aggregation than a full aggregation (Algeria) and that there are even smarter solutions but that this would open the realm of interpolations -:), well, just a suggestion, I am sure, you come up with something clever!

04-spatial-operations.Rmd
```diff
@@ -209,21 +324,60 @@ plot(joined["population_millions"], add = TRUE)
...
+
+<!-- Nearest neighbour analysis -->
+<!-- e.g. two point's datasets (non-overlapping) -->
+<!-- e.g. two point's datasets (overlapping) -->
+<!-- ? topological problems of joining lines/polygons? -->
+<!-- joining different types (e.g. points + polygons = geometry) -> save as GPKG? -->
+<!-- `merge()`; `st_interpolate_aw()` -->
+
+
+<!--### Modifying geometry data; still need to change the corresponding cross-references-->
```
jannes-m: I have changed the heading but I have not changed the corresponding cross-references. I wanted to wait until we agree on a final structure and header naming.

Robinlovelace:
Although... Just seen the build is failing. Please could you take a look before I merge this.

jannes-m:
It was more rearranging than writing, I'd say. Anyways, the checks are passing. When compiling the book, there is a sill warning : `The label(s) modifying-geometry-data not found `. This is because I have changed the name of a heading. But I haven't changed the corresponding cross-references since I wanted to wait until we agree on a final structure and naming (see also one of the comments).

Robinlovelace:
Thanks Jannes, will take a look at these substantial updates. Agree it's still a little up-in-the-air but gradually converging towards a single methods -> applications -> methods -> applications worldview!

jannes-m:
In this case , I was only referring to the structure and section naming of c4 😄. But indeed, we are converging towards a nice chapter structuring!

Robinlovelace:
Haha yes indeed, it's a large PR but I didn't expect you to have written the entire book yet :rofl:

03-attribute-operations.Rmd
Robinlovelace: Great plan, looks good.

PR #87, opened 2017-09-22
suggested io changes

jannes-m:
just started with the review, so there are more commits to come but this way you already can keep track of what I am doing

### DIFF
--- a/05-read-write-plot.Rmd
+++ b/05-read-write-plot.Rmd
@@ -421,20 +421,20 @@ saveWidget(mapview_obj, file="my_interactive_map.html")
 
 1. List and describe three types of vector, raster, and geodatabase formats.
 
-2. Name at least two differences between `read_sf()` and the more well-known function `st_read()`.
+1. Name at least two differences between `read_sf()` and the more well-known function `st_read()`.
 
-3. Read the `cycle_hire_xy.csv` file from the **spData** package (Hint: it is located in the `misc\` folder).
+1. Read the `cycle_hire_xy.csv` file from the **spData** package (Hint: it is located in the `misc\` folder).
 What is a geometry type of the loaded object? 
 
-4. Download the borders of Germany using **rnaturalearth**, and create a new object called `germany_borders`.
+1. Download the borders of Germany using **rnaturalearth**, and create a new object called `germany_borders`.
 Write this new object to a file of the GeoPackage format.
 
-5. Download the global monthly minimum temperature with a temporal resolution of five minutes using the **raster** package.
+1. Download the global monthly minimum temperature with a temporal resolution of five minutes using the **raster** package.
 Extract the June values, and save them to a file named `tmin_june.tif` file (Hint: use `raster::subset()`).
 
-6. Create a static map of Germany's borders, and save it to a PNG file.
+1. Create a static map of Germany's borders, and save it to a PNG file.
 
-7. Create an interactive map using data from the `cycle_hire_xy.csv` file. 
+1. Create an interactive map using data from the `cycle_hire_xy.csv` file. 
 Export this map to a file called `cycle_hire.html`.
 
 [^1]: A list of supported vector formats and theirs options can be found at http://www.gdal.org/ogr_formats.html.
--- a/05-read-write-plot.Rmd
+++ b/05-read-write-plot.Rmd
@@ -11,14 +11,14 @@ library(raster)
 ```
 
 ## Introduction
-
 The previous chapters provided an overview of spatial data classes in R, with a focus on simple features and rasters.
 This chapter is about getting spatial data onto your computer and then, perhaps after processing it with techniques described in this book, back out to the world.
-Geocomputation usually involves reading existing data, and saving it after some processing.
+Before we can even think about processing data, we need to retrieve it, i.e. we need to know which sources are at our disposal (section \@{#retrieving-data}).
+Having data, we need to know how we can import them in our software.
+<!-- so far, I haven't seen a discussion on the pros and cons -->
 Therefore, it is important to be aware of the most often used spatial file formats and their pros and cons (see section \@ref(file-formats)).
 For the same purposes knowing how to efficiently read and write spatial data is crucial (see sections \@ref(data-input) and \@ref(data-output)).
 We also include a section (\@ref(visual-outputs)) on visualization because outputting data in a human (not just computer) readable format enables non-programmers to benefit from your work.
-This if vital if you would like to encourage evidence-based policies.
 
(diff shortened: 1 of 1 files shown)


Robinlovelace:
OK great stuff, many thanks for this. Will 'hold fire' and allow @Nowosad to look over these before merging. This is definitely good stuff!

05-read-write-plot.Rmd
```diff
@@ -2,61 +2,59 @@
 
 The previous chapters provided an overview of spatial data classes in R, with a focus on simple features and rasters.
 This chapter is about getting spatial data onto your computer and then, perhaps after processing it with techniques described in this book, back out to the world.
-Geocomputation usually involves reading existing data, and saving it after processing sections.
-Therefore, it is important be aware of the most often used spatial file format and their strong and weak sides (see section \@ref(file-formats)).
+Geocomputation usually involves reading existing data, and saving it after some processing.
```
Robinlovelace: Yes this is a definitely clearer. Thanks for that.

05-read-write-plot.Rmd
```diff
@@ -421,22 +419,22 @@ saveWidget(mapview_obj, file="my_interactive_map.html")
 
 1. List and describe three types of vector, raster, and geodatabase formats.
 
-1. Name at least differences between `read_sf()` and the more well-known function `st_read()`.
+2. Name at least two differences between `read_sf()` and the more well-known function `st_read()`.
```
Robinlovelace: It does not matter that they are always 1s - Markdown knows what to do: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#lists

05-read-write-plot.Rmd
```diff
@@ -421,22 +419,22 @@ saveWidget(mapview_obj, file="my_interactive_map.html")
...
-1. Create a static map of Germany borders and save it to a PNG file.
+6. Create a static map of Germany's borders, and save it to a PNG file.
 
-1. Create an interactive map using data from the `cycle_hire_xy.csv` file. 
-Export this map to the HTML file called `cycle_hire.html`.
+7. Create an interactive map using data from the `cycle_hire_xy.csv` file. 
+Export this map to a file called `cycle_hire.html`.
 
-[^1]: A list of supported vector formats and theirs options could be found at http://www.gdal.org/ogr_formats.html.
+[^1]: A list of supported vector formats and theirs options can be found at http://www.gdal.org/ogr_formats.html.
```
Robinlovelace: Yes this is definitely right - alongside your other changes these are great improvements.

05-read-write-plot.Rmd
```diff
@@ -1,62 +1,167 @@
...
 
 I/O is short for "input/output" which means, in plain English, "reading and writing data".
-We use the acronym instead of plain English not to confuse you or to make chapter names short, but because that's the term used in computer science and it is useful to think of data import and export from a computing perspective.^[
+We use the acronym instead of plain English not to confuse you or to make chapter names short, but because that is the term used in computer science, and it is useful to think of data import and export from a computing perspective.^[
 Concepts such as computational efficiency, hard disk space and 'idempotence' are useful when thinking about reading and writing geographic datasets, which can become large and difficult to handle.
 Loading/saving data is yet another way of saying the same thing.
 ]
 
-## File formats
+## Retrieving open data {#retrieving-data}
```
jannes-m: I would start with retrieving data since this is the natural first step before doing anything else such as importing it.

05-read-write-plot.Rmd
```diff
@@ -1,62 +1,167 @@
...
+Nowadays, a vast amount of spatial data is available on the internet.
```
jannes-m: Here, I have just provided a little bit of context (geolibraries, geoportals and corresponding examples).

05-read-write-plot.Rmd
```diff
@@ -150,106 +253,22 @@ raster_filepath = system.file("raster/srtm.tif", package = "spDataLarge")
...
-single_layer_band3 = raster(raster_multilayer_filepath, band = 3)
-```
-
-The whole multilayer raster file is read by `brick()`.
-
-```{r}
-multilayer_layer_brick = brick(raster_multilayer_filepath)
-```
-
-Finally, many single and multilayer rasters could be read together using the `stack()` function:
```
jannes-m: I deleted this part because we showed similar code in c2. But still there is a problem, in c2 we are already presenting how to read in raster data using raster, stack and brick. And then explain what the differences are. Hence, I would suggest to just show in c2 how to manually create a raster and how to stack two rasters while referring to c5 for a more detailed explanation regarding raster stacks and bricks (i.e. move great parts of the raster classes section to c5). What do you think?

05-read-write-plot.Rmd
```diff
@@ -407,36 +428,36 @@ save_tmap(tm  = tmap_obj, filename = "lifeExp_tmap.png")
 <!-- a = plot(world["lifeExp"]) -->
 <!-- ``` -->
 
-On the other hand, an interactive map could be usually saved as an HTML file. 
-This can be done with the `saveWidget()` function from the `htmlwidget` package:
+On the other hand, on can save interactive maps as an HTML file using the `saveWidget()` function from the `htmlwidgets` package:
 
+<!-- example doesn't work, problem with colors I guess -->
```
jannes-m: the mapview example doen't work, pls fix

05-read-write-plot.Rmd
```diff
@@ -150,109 +260,26 @@ raster_filepath = system.file("raster/srtm.tif", package = "spDataLarge")
...
-world_raw_filepath = system.file("shapes/world.gpkg", package = "spData")
-world_raw = st_read(world_raw_filepath)
-```
-
-More information on getting data using R packages could be found in [section 5.5](https://csgillespie.github.io/efficientR/input-output.html#download) and [section 5.6](https://csgillespie.github.io/efficientR/input-output.html#accessing-data-stored-in-packages) of @gillespie_efficient_2016.
+Please refer to section \@ref(raster-classes) for information on the difference between raster stacks and bricks.
 
 ## Data output (O) {#data-output}
 
+<!--maybe we can come up withan intro which is a bit more compelling-->
```
jannes-m: I think we can write an intro which is a bit more compelling. I have to run now, so maybe you can come up with sth. or else I will think about it later on.

jannes-m:
For the moment, I am done with reviewing c5, so feel free to merge.

05-read-write-plot.Rmd
Nowosad: That's a bug in mapview. It is already fixed in the dev version on gh, so it should be on CRAN soon.

PR #92, opened 2017-10-12
Location analysis rev1

Robinlovelace:


Robinlovelace:
This looks like another excellent applied chapter in the making. It's still not reproducible I don't think - the Rmd file crashes on my computer with the following message:

```
Quitting from lines 212-233 (08-location.Rmd) 
Error in select.list(points, osm_points) : 
  select.list() cannot be used non-interactively
Calls: local ... eval -> lapply -> FUN -> <Anonymous> -> select.list
Execution halted
Error in Rscript_render(f, render_args, render_meta) : 
  Failed to compile 08-location.Rmd
Calls: <Anonymous> ... render_new_session -> tryCatch -> tryCatchList -> Rscript_render
Execution halted
```

Plus not working on Travis. @Nowosad could we put the example data in **spDataLarge** (seems over a couple of MB).

Robinlovelace:
But overall I suggest you just merge this and keep plugging (or should I say pedalling ; ) away at it.

jannes-m:
@Robinlovelace Thanks for the great input, really appreciate it! I have incorporated your suggestions manually due to unresolvable merge conflicts (not the best way to go, but sometimes easier...). So I suggest that you delete this branch.
I would rather not download the German census raster and the osmdata every time we commit something. So I suggest to either set eval to FALSE (current approach) or to cache the chunk. What do you prefer?

I have also updated the osmdata code chunk. Now it works again. Let me know if you still encounter trouble ( @Nowosad ).

Robinlovelace:
No problem and thanks for the manual merge! Yes I agree we shouldn't have to pull down the census on each build but even more importantly it should build. What did you think of the suggestion of putting the data needed for this in one of Jakub's data pkgs?

PR #94, opened 2017-10-12
Transport chapter work in progress

Robinlovelace:


jannes-m:
I think this is quite an interesting application. Agent simulation modeling might be a bit advanced but I might be wrong (haven't really used simulations so far, I just know what they are supposed to do). I am also interested in your routing approach, i.e. which package are you going to use. 

Overall, I think it would be great if we managed to point out the strengths of R in our our use case chapters (e.g., statistical power, visualization power, great interfaces, readily available packages).

I have just come across this [blog](https://insightr.wordpress.com/2017/10/13/uber-assignment-with-lpsolve/#comments). Seems to be related. Do you also use **lpSolve**?

Robinlovelace:
Thanks Jakub for the typo fix and Jannes for the comments. Yes I think the agent-based modelling part will be a description rather than code: it could already get very long!

I plan to use **dodgr** by @mpadge for the routing. He'll be here in Leeds in a couple of weeks to help explain that to an introductory audience, beyond his excellent **vignette**: https://gmost.github.io/dodgr/articles/dodgr.html

The **lpsolve** looks to solve a wider problem: the travelling salesman. Not planning to cover that but, like the ABM stuff would be good to do so in future chapters if we get the time - would be really fun the ABM stuff - any ecological applications you know of? Was thinking it could be a chapter.

jannes-m:
As I said I haven't used simulations but they have their merits for example for wildfire, natural hazard, earthquake (building), escape route simulations.

**dodgr**, wow, interesting, didn't know of this package, I have used **osmar** and **igraph** for shortest route calculations.

PR #115, opened 2017-11-22
Geomarketing rev rl

Robinlovelace:
This is my review - looking great.

### DIFF
--- a/08-location.Rmd
+++ b/08-location.Rmd
@@ -14,10 +14,10 @@ library(tidyverse)
 ## Introduction
 
 This chapter demonstrates how the skills learned in the previous chapters can be applied to location analysis.
-Location analysis is a broad field of research and (largely) commercial application, the aim of which is to decide the optimal location for interventions.
+Geomarketing (also called) location analysis is a broad field of research and (largely) commercial application, the aim of which is usually to decide the optimal location for new services.
 A typical example is where to locate a new shop: there are typically hundreds of possible locations in a city and location analysis can be used to decide where the shop will attract most visitors and, ultimately, make most profit.
 There are also many non-commercial applications that can use the technique for public benefit.
-A good example is where to locate new health services benefiting as many people as possible [@tomintz_geography_2008].
+A good example is where to locate new health services [@tomintz_geography_2008].
 
 Ecological concepts can be useful in understanding location analysis.
 Animals have needs which they tend to meet by finding the optimal location based on variables that change over space (called gradients in the ecological literature --- see chapter xx).
--- a/08-location.Rmd
+++ b/08-location.Rmd
@@ -19,23 +19,23 @@ A typical example is where to locate a new shop: there are typically hundreds of
 There are also many non-commercial applications that can use the technique for public benefit.
 A good example is where to locate new health services [@tomintz_geography_2008].
 
-Ecological concepts can be useful in understanding location analysis.
-Animals have needs which they tend to meet by finding the optimal location based on variables that change over space (called gradients in the ecological literature --- see chapter xx).
+Geomarketing is focused on people and where they are likely to spend their time and other resources.
+However, it has several links with ecology:
+animals have needs which can be best met in optimal locations based on variables that change over space (called gradients in the ecological literature --- see chapter xx).
 <!-- add reference!! -->
-A polar bear, for example, may prefer northern latitudinal regions where the ice is persistent throughout the summer, and where food (seals and sea lions) and shelter are plentiful.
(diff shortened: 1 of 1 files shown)


Robinlovelace:
General pointers - please could you replace **plyr** with **dplyr** if possible. I've checked out if there's a way to do that and think there is: https://stackoverflow.com/questions/30597522/what-is-the-dplyr-equivalent-of-plyrldplytapply-in-r

Robinlovelace:
Overall the work is excellent though. If you're happy to merge this PR and polish it a bit I think we should merge it and get it building on the main book website ASAP. Then suggest you, @Nowosad and I make a few additional tweaks on it. Sound like a plan?

Robinlovelace:
Also was thinking at points the code chunks are a little long but that may be inevitable. However if you see any opportunities to reduce the length of the code please go for them!

jannes-m:
@Robinlovelace Thanks for the review!! Will take care of the issues you have pointed out. But I don't think there is a way to replace **plyr** with **dplyr**. The latter is soleley written for data.frame-like structures. When using `lists`, **dplyr** is the wrong candidate. But if you want to get rid off `ldply()`, we could use `lapply()` and then do `do.call(rbind, ...)`. But maybe it's worth mentioning that there is a split-apply approach, and that not everything in R is always a data.frame.

Robinlovelace:
I would go with `lapply()` and (if it works on the data) `dplyr::bind_rows()` which binds lists together.

Sorry if it seems I am 'binded' to the **tidyverse**: I'm not. Just wanting to avoid the confusion of having **dplyr** and its (partial) predecessor **plyr** in the same book.

Regarding next steps are you happy with the merge this -> tweak -> merge with master -> polish plan?

jannes-m:
Hey Robin,
well, we can go with `lapply()`. However, I am not quite sure regarding the reasoning. We shouldn't use **plyr** because of **dplyr**. I don't see any confusion there. d(ataframe)plyr is used for `data.frames` but pretty much useless when it comes to other data types such as vectors, lists, etc. because it wasn't build for it. **plyr** is just a more generalized form of **dplyr**, and in this sense not a predecessor. Therefore, it seems strange that **plyr** was released prior to **dplyr** but well... I hope I haven't bored you with explanations you are already aware of. Anyway, maybe something to discuss tomorrow. And yes, we can merge, and I make the remaining adjustments in the master branch.

Robinlovelace:
No all good but I think `purrr::map*()` functions would be the recommended solution by the package's developer @hadley these days. Not being a heavy user of either I cannot say for sure but I'm confident that ditching **plyr** is a good thing for our readers. It seems it was last updated on CRAN over a year ago: https://cran.r-project.org/web/packages/plyr/index.html

Hadley any feedback on this welcome. I note that in r4ds there is no mention for **plyr** but plenty on **purrr** which implies to me its developer sees it as a better solution. There's a good section on **purrr** here but suggest we stick with good old fashioned `lapply()` for now to avoid #pkgoverload!

Nowosad:
Hi @jannes-m @Robinlovelace - I'm here just for the plyr/dplyr/purrr discussion;)
I agree with @Robinlovelace - we should not use plyr.
 In my work I often use purrr and it is really good (yes, map is almost the same as lapply, but there are many more nice things, such as map_*, pmap, imap, etc). That being said, base R solution is ok in this case.

jannes-m:
I am a bit irritated by how you pass functions in **purrr**, you use `~` instead of function(x). This might be a good thing when doing multiple models, in other cases it is somewhat confusing. Here, a silly example:

```
metros = c("Berlin", "Hamburg", "Munich")
lapply(metros, function(metro) {
  print(metro)
}
```
and with **purrr**:

```
map(metros, ~ print(.))
```

It's short but I think it is not really clear what is going on. It becomes even more confusing if you start using `.$`.  But ok, if the two of you like it better than **plyr**, then I will use either **purrr** or `lapply()`. It's really not a big issue, we just have to agree on something.

Nowosad:
You don't need to use `~` with purrr:
```
#1
map(metros, ~print(.)) 
#2
map(metros, print) 
#3
map(metros, function(metro) { 
  print(metro)
})
```

jannes-m:
Ah, see, now I like it much better! Perfect, thanks for the clarification. So let's purrr!

Robinlovelace:
Also **purrr**ing does not leed to #pkgoverload because it's already in the **tidyverse**. Thanks for wading into this sticky debate Jakub and shedding more light than heat onto it!

PR #129, opened 2017-12-10
Geomarketing rev rl2

Robinlovelace:


07-location.Rmd
```diff
@@ -18,14 +18,15 @@ library(tidyverse)
...
-There are also many non-commercial applications that can use the technique for public benefit.
-A good example is where to locate new health services [@tomintz_geography_2008].
+This chapter demonstrates how the skills learned in Part I can be applied to a particular domain: location analysis (also called geomarketing).
+This is a broad field of research and commercial application, the aim of which is usually to decide the optimal location for new services.
+A typical example is where to locate a new shop.
+The aim here is to attract most visitors and, ultimately, make most profit.
+There are also many non-commercial applications that can use the technique for public benefit, for example where to locate new health services [@tomintz_geography_2008].
 
-Geomarketing is focused on people and where they are likely to spend their time and other resources.
-However, it has several links with ecology: animals and plants have needs which can be best met in optimal locations based on variables that change over space (called gradients in the ecological literature --- see e.g., @muenchow_review_2017<!--and chapter xx-->).
```
Robinlovelace: Making this shorter because the term introduced, gradients, does not re-appear in the chapter. Easier to read without the brackets.

07-location.Rmd
```diff
@@ -42,11 +42,11 @@ Typical research questions include:
...
 
-## Case study
+## Case study: bike shops in Germany
 
-Imagine your are launching a cycling company and would like to open shops across urban areas of Germany.
-Imagine further that a survey you conducted indicates following target group: males between 20-40 years old who live alone or with just one person (single households, not families).
-Additionally, you are in the lucky position to have sufficient capital to open a number of shops.
+Imagine you are starting a chain of bike shops in Germany.
+A survey suggests that single males between aged 20 to 40 are most likely to buy your products: this is the *target audience*.
+You are in the lucky position to have sufficient capital to open a number of shops.
```
Robinlovelace: How many? I think it's important to provide a ball-park figure so the reader has a sense of the size of the challenge.

07-location.Rmd
jannes-m: Ok

Robinlovelace: Yes - that's a vital point and set-up the eco chapter perfectly. The thinking is that introducing a new concept that is not used again in the chapter will dilute the point, not strengthen it, at this stage. We could talk about gradients of demand later in the chapter and re-instate it though but, re-reading it, I think there's already a lot to get one's head around. Hence the mission to simplify things and rely on links to further resources for advanced readers. Adding the link to the eco section will help with that so thanks for adding that in commented form.

jannes-m: ok, agreed!

Robinlovelace:
Thanks for fixing the merge conflict Jannes!

PR #133, opened 2017-12-20
Reviewing c4 (jm)

jannes-m:
Hey Robin, 
I have reviewed the intro and the vector part of c4, great work! I hope to have a look at the raster part tomorrow or the day after tomorrow.


04-spatial-operations.Rmd
```diff
@@ -60,6 +65,8 @@ canterbury = nz %>% filter(REGC2017_NAME == "Canterbury Region")
 canterbury_height = nz_height[canterbury, ]
 ```
 
+<!-- maybe we should add to the caption that many of the high points are close to each other which is why one gets the impression in NZ there were only ca. 15 high points, but nz_height contains 101 points -->
```
jannes-m: maybe we should add to the caption that many of the high points are close to each other which is why one gets the impression in NZ there were only ca. 15 high points, but nz_height contains 101 points

04-spatial-operations.Rmd
```diff
@@ -82,6 +89,7 @@ These determine the type of spatial relationship that features in the target obj
 *Intersects* is the default spatial subsetting operator, a default that returns `TRUE` for many types of spatial relations, including *touches*, *crosses* and *is within*.
 These alternative spatial operators can be specified with the `op =` argument, as illustrated below (try plotting the result or skip to section \@ref(topological-relations) to find out what this does):
 
+<!-- any reason why you use , , instead of nz_height[canterbury, , op = st_disjoint]. Additionally, I think it would be more satisfying for the reader to already state here that st_disjoint is the opposite of st_intersects  -->
```
jannes-m: Any reason why you use , , instead of nz_height[canterbury, , op = st_disjoint]. Additionally, I think it would be more satisfying for the reader to already state here that st_disjoint is the opposite of st_intersects.
Pls note also a remark below where I will suggest to switch places for sections subsetting and topological relations

04-spatial-operations.Rmd
```diff
@@ -140,13 +148,14 @@ For further discussion of this decision, and some controversy, see the (closed)
 
 But what is `sel`?
 It is not, as one might imagine, a `logical` vector.
+<!-- What do you mean by two-dimensional? One can misunderstand this and believe that you are talking about 2 columns. Or are you already referring to sel_matrix? -->
```
jannes-m: What do you mean by two-dimensional? One can misunderstand this and believe that you are talking about 2 columns. Or are you already referring to `sel_matrix`?

04-spatial-operations.Rmd
```diff
@@ -168,7 +177,7 @@ typeof(sel_matrix)
 dim(sel_matrix)
 ```
 
-
+<!-- What do you think about placing this section before the spatial subsetting section since it sets the scene for subsetting, join, etc. -->
```
jannes-m: What do you think about placing the topological section before the spatial subsetting section since it sets the scene for subsetting, join, etc.

04-spatial-operations.Rmd
```diff
@@ -404,7 +413,7 @@ joined = st_join(x = asia, y = urb) %>%
   na.omit()
 ```
 
-
+<!-- The yellow point is hardly recognizable, maybe use also the border parameter? -->
```
jannes-m: The yellow point is hardly recognizable in the figure (Tokio), maybe use also the border parameter?

04-spatial-operations.Rmd
```diff
@@ -520,11 +529,13 @@ plot(z["capacity"])
 <!--### Modifying geometry data; still need to change the corresponding cross-references-->
 
 The result of this join has used a spatial operation to change the attribute data associated with simple features but the geometry associated with each feature has remained unchanged.
+<!-- no longer true, that's why I would remove the sentence
```
jannes-m: no longer true (these operations are now in c5), that's why I would remove the sentence here

04-spatial-operations.Rmd
```diff
@@ -559,18 +570,20 @@ The resulting `nz_avheight` objects have the same geometry as the aggregating ob
 Note that regions containing no points have an associated `elevation` value of `NA`.
 For aggregating operations which also create new geometries, see section \@ref(geometry-unions).
 
+<!-- One should avoid to have a single subsection, here 4.2.5.1. I would delete the heading, then spatial congruence just becomes another paragraph of spatial aggregation -->
```
jannes-m: One should avoid to have a single subsection, here 4.2.5.1. I would delete the heading, then spatial congruence just becomes another paragraph of spatial aggregation

04-spatial-operations.Rmd
```diff
@@ -559,18 +570,20 @@ The resulting `nz_avheight` objects have the same geometry as the aggregating ob
...
 
 *Incongruent* aggregating objects, by contrast, do not share common borders with the target [@qiu_development_2012].
 This is problematic for spatial aggregation (and other spatial operations) illustrated in Figure \@ref(fig:areal-example).
 Areal interpolation can help to alleviate this issue.
 It helps to transfer data from one set of areal units to another.
-A number of algorithms have been developed for areal interpolation, including area weighted and pycnophylactic interpolation methods task [@tobler_smooth_1979].
+A number of algorithms have been developed for areal interpolation, including area weighted and pycnophylactic interpolation methods [@tobler_smooth_1979].
 
+<!-- somehow the incongruent borders are hardly to discern in the right figure, maybe increasing the figure size could help -->
```
jannes-m: somehow the incongruent borders are hardly to discern in the right figure, maybe increasing the figure size could help (though the figure is really nice!)

Robinlovelace:
Many thanks Jannes for this comprehensive review. Having looked at the changes I think they make the chapter much stronger. I'll merge now and reply to your comments making further changes as I go and provide further feedback as appropriate.

04-spatial-operations.Rmd
```diff
@@ -26,16 +26,22 @@ library(spData)
...
 This chapter shows how spatial objects can be modified in a multitude of ways based on their location and shape.
 The content builds on the previous chapter because many spatial operations have a non-spatial (attribute) equivalent.
-Spatial operations on *vector* objects include spatial subsetting (covered in section \@ref(spatial-subsetting)) and spatial joining (section \@ref(spatial-joining)).
-This content builds on section \@ref(vector-attribute-manipulation) in the previous chapter.
-Spatial operations on *rasters* include merging and subsetting, covered in section \@ref(spatial-ras).
-
-The chapter also introduces new concepts that are unique to spatial data.
-A variety of *topological relations* can be used to subset/join vector geometries (by default **sf** uses the catch-all *intersects* but other relations such as *within* can be very useful), a topic that is explored in section \@ref(topological-relations).
-Spatial operations on raster datasets involve *map algebra* (covered in sections \@ref(map-algebra) to \@ref(global-operations-and-distances)) and raster merging (covered in sections \@ref(merging-rasters)).
+This is especially true for *vector* operations.
```
Robinlovelace: Good point

04-spatial-operations.Rmd
```diff
@@ -26,16 +26,22 @@ library(spData)
...
+For instance, we will extend the basics on vector attribute manipulation (covered in section \@ref(vector-attribute-manipulation)) with its spatial counterpart, that is spatial subsetting (covered in section \@ref(spatial-subsetting)) and spatial joining (section \@ref(spatial-joining)).
+Imagine you want to add an attribute of a road (such as the road number or the speed limit) to a municipality polygon.
+Equally, you might want to select all roads of a specific municipality.
+Were we in possession of an ID this would simply repeat the stuff introduced in the previous chapter.
```
Robinlovelace: I know what you mean but will reword - I am in possession of several IDs!

04-spatial-operations.Rmd
Robinlovelace: Agree that it's more satisfying to *know* what it does (although it's good to provide incentives to keep reading!).

I've updated the line in any case so it now reads:

```
These alternative spatial operators can be specified with the `op =` argument.
This is demonstrated in the following command which returns the opposite of `st_intersect()`, points that do not intersect with Cantebury (see in section \@ref(topological-relations)):
```

However, I'm not sure what you mean by the following:

> any reason why you use , , instead of nz_height[canterbury, , op = st_disjoint].

I didn't use `, , ` but:

```r
nz_height[canterbury, , op = st_disjoint]
```

which is what you're recommended, unless I'm missing something or you made a typo in that comment.

jannes-m: 😄 , my mistake. I was wondering, why you use 
`nz_height[canterbury, , op = st_disjoint]`
instead of:
`nz_height[canterbury, op = st_disjoint]`

Robinlovelace: To show that the `[` operator expects a third argument. Is there a strong reason to remove it? I think it risks people thinking the second argument in `[` for spatial objects is `op`, which is untrue: that is a good reason for it to stay 'as is'.

04-spatial-operations.Rmd
```diff
@@ -26,17 +26,22 @@ library(spData)
...
-
-The chapter also introduces new concepts that are unique to spatial data.
-A variety of *topological relations* can be used to subset/join vector geometries (by default **sf** uses the catch-all *intersects* but other relations such as *within* can be very useful), a topic that is explored in section \@ref(topological-relations).
-Spatial operations on raster datasets involve *map algebra* (covered in sections \@ref(map-algebra) to \@ref(global-operations-and-distances)) and raster merging (covered in sections \@ref(merging-rasters)).
-
+This is especially true for *vector* operations.
+For instance, we will extend the basics on vector attribute manipulation (covered in section \@ref(vector-attribute-manipulation)) with its spatial counterpart, namely spatial subsetting (covered in section \@ref(spatial-subsetting)) and spatial joining (section \@ref(spatial-joining)).
+Imagine you want to add an attribute of a road (such as the road number or the speed limit) to a municipality polygon.
+Equally, you might want to select all roads of a specific municipality.
+Were we in possession of an ID this would simply repeat the stuff introduced in the previous chapter.
```
Robinlovelace: "simply repeat the stuff" sounds informal. I've reworded it.

04-spatial-operations.Rmd
```diff
@@ -26,17 +26,22 @@ library(spData)
...
-Spatial operations on raster datasets involve *map algebra* (covered in sections \@ref(map-algebra) to \@ref(global-operations-and-distances)) and raster merging (covered in sections \@ref(merging-rasters)).
-
+This is especially true for *vector* operations.
+For instance, we will extend the basics on vector attribute manipulation (covered in section \@ref(vector-attribute-manipulation)) with its spatial counterpart, namely spatial subsetting (covered in section \@ref(spatial-subsetting)) and spatial joining (section \@ref(spatial-joining)).
+Imagine you want to add an attribute of a road (such as the road number or the speed limit) to a municipality polygon.
+Equally, you might want to select all roads of a specific municipality.
+Were we in possession of an ID this would simply repeat the stuff introduced in the previous chapter. 
+But if we are missing an ID, then we can still solve these questions using the geographical location of the objects in question.
+But then we might face further questions such as: Should the road completely fall inside the polygon or is it enough if the road simply crosses the municipality or is it even okay if it is within a certain distance of it.
```
Robinlovelace: Again overly informal language: there's a whole article urging teachers not to use words like 'simply', especially for programming: https://medium.freecodecamp.org/one-does-not-simply-learn-to-code-f25bacdc5b62

Reworded - see what you think and let me know if there's any other issues. I agree with the direction of your rewording though, great to have a concrete example of roads (makes me think of more content in Part II or IV).

04-spatial-operations.Rmd
jannes-m: Just looks a bit strange, and therefore might be confusing. On the other hand I like that you make clear that all columns should be selected. However this would also imply to always use a comma prior to a column name in column subsetting, e.g., `nz_height[, "height"]` instead of `nz_height["height"]` as is sometimes done in our code chunks.

04-spatial-operations.Rmd
jannes-m: You are right, thanks.

04-spatial-operations.Rmd
jannes-m: I like your rewrite, and yes, this sentence, too, was too informal, sorry. And we don't have to use "simply". But here simply didn't refer to coding but if a road crosses a polygon. Nevertheless, we can omit the word. 
Regarding the "simply" post, I understand what the author wants to say. On the other hand, I would also point out that coding is not rocket science, and of course, everybody can learn it. But like with any other language or skill it takes time to become proficient.
Still, I will try to omit "simply" in the future.

04-spatial-operations.Rmd
Robinlovelace: That's a good point about `obj[col]` vs 'obj[, col]`.

There is an important difference for data frames, as illustrated by:

``` r
head(cars["dist"])
#>   dist
#> 1    2
#> 2   10
#> 3    4
#> 4   22
#> 5   16
#> 6   10
head(cars[, "dist"])
#> [1]  2 10  4 22 16 10
```

This makes me think of a more important improvement: mention that `obj[col]` is the same as `obj[,col]` for `sf` objects but not `data.frame`s in c3. Up for giving that a bash?

I think mentioning that the empty `, ,` part means 'all columns' is the best solution - `obj[col]` is frequently used and widely understood but `obj[sp_ob, , op = st_*]` is not so should be spelt out explicitly to avoid the much greater risk of causing confusion.

jannes-m: Sounds good. And yes, I can mention this in c3.

One advantage of **dplyr** is that it always returns a dataframe object. In base R this is not the case as shown by your `cars` example. However, `cars[, "dist", drop = FALSE]` would also return a dataframe. In the case of sf-objects a dataframe object is returned when you only select a single attribute, I guess, because of the sticky geometry column, i.e., in fact at least two columns are returned which avoids the automatic coercion into the lowest possible dimension as done by  `[`.

Robinlovelace: Interesting points worth mentioning. A good candidate for an rmdnote?

04-spatial-operations.Rmd
Robinlovelace: Job done.

04-spatial-operations.Rmd
Robinlovelace: The explanation of the empty `, ,` bit has now been added. Tag me if/when you add a not in c3 and I'll take a look.

04-spatial-operations.Rmd
Robinlovelace: Solved now - thanks for the comment, there were defo issues here.

04-spatial-operations.Rmd
Robinlovelace: Good point. Sorted with a rainbow pal. One question: how to make other pallettes (e.g. `viridis::*` ones) work?

04-spatial-operations.Rmd
Robinlovelace: Thanks - I've also tidied up by removing this comment now!

04-spatial-operations.Rmd
```diff
@@ -534,7 +545,7 @@ This is a good example of spatial aggregation: it is the geometry of the source
 This is illustrated using the base `aggregate()` function below:
 
 ```{r}
-nz_avheight = aggregate(nz_height, nz, FUN = mean)
+nz_avheight = aggregate(x = nz_height, nz, FUN = mean)
```
Robinlovelace: As a stylistic preference, I've read that it's conventional for the first argument *not* to be stated and I think this is generally the case. However it does not harm and sometimes it's good to be explicit (thinking of the `, ,` debate) so leaving this as is. Thanks.

04-spatial-operations.Rmd
Robinlovelace: Agreed. Job done.

04-spatial-operations.Rmd
jannes-m: Ah ok.The reason I did so here, was because in the text you explicitly mentioned x and y.

Robinlovelace: That's a v. good reason!

04-spatial-operations.Rmd
Robinlovelace: Good point - much better now I think. Made it larger and added a border: 
image

Robinlovelace: In this commit: https://github.com/Robinlovelace/geocompr/commit/b7f9d2b50cbf34fdd511ae73f38720541cb2d062

04-spatial-operations.Rmd
Robinlovelace: Fair point - simply has its place and agree that it was referring to a road not code.

04-spatial-operations.Rmd
Robinlovelace: Nudge.

04-spatial-operations.Rmd
jannes-m: 👍

PR #138, opened 2018-01-11
Reviewing transport chapter

jannes-m:
So far I have been reviewed until 8.4. Still, I already have some suggestions.


08-transport.Rmd
```diff
@@ -12,7 +12,9 @@ library(tidyverse)
 
 ## Introduction
 
+<!-- I would tone down, why should geographic space be more tangible in transport than in remote sensing, location analysis, ecological modeling, etc.? -->
```
jannes-m: I would tone down, why should geographic space be more tangible in transport than in remote sensing, location analysis, ecological modeling, etc.

08-transport.Rmd
```diff
@@ -12,7 +12,9 @@ library(tidyverse)
 
 ## Introduction
 
+<!-- I would tone down, why should geographic space be more tangible in transport than in remote sensing, location analysis, ecological modeling, etc.? --> 
 In no other sector is geographic space more tangible than transport.
+<!-- I am not 100% sure but I think the first law of geography was the basis for spatial dependence, spatial autocorrelation, and therefore the foundation for spatial interpolation. Not sure how 'moving' relates to this but this just might be because I am not a transportation guy. -->
```
jannes-m: I am not 100% sure but I think the first law of geography was the basis for spatial dependence, spatial autocorrelation, and therefore the foundation for spatial interpolation. Not sure how 'moving' relates to this but this just might be because I am not a transportation guy.

08-transport.Rmd
```diff
@@ -28,36 +30,41 @@ It is therefore unsurprising that transport researchers have long turned to geoc
 This chapter provides an introduction to geographic analysis of transport systems.
 We will explore how movement patterns can be understood at multiple geographic levels, including:
 
-- **Areal units**: transport can be understood simply in terms of zonal aggregates such as the main mode and average distance of trips made people living in a particular zone.
+<!-- What is the main mode? I don't know the term. -->
```
jannes-m: What is the main mode? I don't know the term. Reading on I understood what you meant but maybe a little clarification here would be helpful.

08-transport.Rmd
```diff
@@ -28,36 +30,41 @@ It is therefore unsurprising that transport researchers have long turned to geoc
...
 - **Routes**: these are circuitous (non-straight) routes, typically representing the 'optimal' path along the route network between origins and destinations along the desire lines defined in the previous bullet point.
-- **Nodes**: these are points in the transport system that can represent common origins and destinations (e.g. with one centroid per zone) and public transport stations such as bus stops and rail stations.
+- **Nodes**: these are points in the transport system that can represent common origins and destinations (e.g., with one centroid per zone) and public transport stations such as bus stops and rail stations.
 - **Route networks**: these represent the system of roads, paths and other linear features in an area. They can be represented as purely geographic entities or as a graph.
+<!-- Not sure but maybe a word is missing, to me values representing 'flow' sounds a bit awkward (but then again I am not a native speaker --> 
 Their features are segments which can be assigned values representing 'flow', the number of people expected to use a particular street or path.
```
jannes-m: Not sure but maybe a word is missing, to me values representing 'flow' sounds a bit awkward (but then again I am not a native speaker

08-transport.Rmd
```diff
@@ -28,36 +30,41 @@ It is therefore unsurprising that transport researchers have long turned to geoc
...
 
 The lowest but hardest-to-model level in transport systems is **agents**, mobile entities like you and me.
-Some models built using software such as [MATSim](http://www.matsim.org/) can effectively capture the dynamics of transport systems using an agent-based modelling (ABM) approach at high spatial and temporal resolution.
+Some models can effectively capture the dynamics of transport systems using an agent-based modelling (ABM) approach at high spatial and temporal resolution, e.g., [MATSim models](http://www.matsim.org/).
+<!-- is that true? More inaccessible than geocomputation or geostats with R? -->
 These tend to be complex and computationally demanding [@horni_multi-agent_2016], meaning they are inaccessible to most people.
```
jannes-m: Is that true? More inaccessible than geocomputation or geostats with R?

08-transport.Rmd
```diff
@@ -28,36 +30,41 @@ It is therefore unsurprising that transport researchers have long turned to geoc
...
 <!-- Idea: make it about reducing CO2 emissions instead. Thoughts? + Multi-model - more complex -->
 
 ## A case study of Bristol {#bris-case}
 
-The case study used for this chapter is a city on the west of England, around 30 km East of the Welsh capital Cardiff.
-An overview of the region's transport network is illustrated in Figure \@ref(fig:bristol), which shows a diversity of transport infrastructure, for cycling, public transport, and private motor vehicle.
+The case study used for this chapter is located in Bristol - a city in the west of England, around 30 km East of the Welsh capital Cardiff.
+An overview of the region's transport network is illustrated in Figure \@ref(fig:bristol), which shows a diversity of transport infrastructure, for cycling, public transport, and private motor vehicles.
 
+<!-- there should be a legend for the study borders (blue lines) as well. I would also reduce the length of the caption since it is basically repeating the content of the legend. -->
```
jannes-m: there should be a legend for the study borders (blue lines) as well. I would also reduce the length of the caption since it is basically repeating the content of the legend

08-transport.Rmd
```diff
@@ -102,12 +109,13 @@ View(cw0103)
...
+There is an ambitious [plan](http://www.cyclingweekly.com/news/interview-bristols-mayor-george-ferguson-24114) to double the share of cycling by 2020.
 
-In this policy context the aim of this chapter, beyond demonstrating how geocomputation with R can be used to support sustainable transport planning, is to provide evidence for decision-makers in Bristol to decide how best to increase the mode share of walking and cycling in particular in the city.
+In this policy context the aim of this chapter, beyond demonstrating how geocomputation with R can be used to support sustainable transport planning, is to provide evidence for decision-makers in Bristol to decide how best to increase the share of walking and cycling in particular in the city.
 This high-level aim will be met via the following objectives:
 
 - Describe the geographical pattern of transport behavior in the city.
+<!-- potential bottlenecks? you mean where cycling/walking cannot be further increased? --> 
 - Identify key nodes and potential bottlenecks where walking and cycling potential is likely to be highest.
```
jannes-m: Potential bottlenecks? You mean where cycling/walking cannot be further increased?

08-transport.Rmd
```diff
@@ -117,24 +125,24 @@ These zone-level data are small but often vital for gaining a basic understandin
 
 ## Transport zones
 
-Although transport systems are primarily based on linear features and nodes --- including pathways and stations --- it often makes sense to start with areal data.
-Three types of zone will typically be of particular interest: the study region, origin zones (typically residential areas), and destination zones (which are often the same as the origin zones).
+Although transport systems are primarily based on linear features and nodes --- including pathways and stations --- it often makes sense to start with areal data <!-- source and/or example -->.
```
jannes-m: Add reference and/or example.

08-transport.Rmd
```diff
@@ -117,24 +125,24 @@ These zone-level data are small but often vital for gaining a basic understandin
...
+Three zone types will typically be of particular interest: the study region, origin zones (typically residential areas), and destination zones (which are often the same as the origin zones). <!-- this implies people are going from their origin to the origin possibly via a detour (which might be work/supermarket/etc.-->
```
jannes-m: After reading on I got what you were referring to but here it is a bit confusing to say that origin and destination zones are the sames when saying that origin zones are typically residential areas. Do you know what I mean?

08-transport.Rmd
```diff
@@ -117,24 +125,24 @@ These zone-level data are small but often vital for gaining a basic understandin
...
+In this case the command returns a detailed polygon representing the official boundary of Bristol (see the inner blue boundary in Figure \@ref(fig:bristol)).
+A word of warning: First, OSM boundaries as returned by OSM are not necessarily the official boundaries provided by local authorities.
+Secondly, even if OSM returns the official boundaries, these still may not be appropriate for transport research because they bear little relation to where people travel.
 
-Travel to Work Areas (TTWAs) overcome this issue by creating a zoning system analogous to hydrological watersheds.
-TTWAs were first defined as contiguous zones within which 75% of the population travels work [@coombes_efficient_1986], and this is the definition used in this chapter.
+Travel to Work Areas (TTWAs) overcome the second issue by creating a zoning system analogous to hydrological watersheds.
+TTWAs were first defined as contiguous zones within which 75% of the population travels to work [@coombes_efficient_1986], and this is the definition used in this chapter.
```
jannes-m: Pls check if `travels to work` is what you mean!

08-transport.Rmd
```diff
@@ -143,10 +151,11 @@ region_ttwa = readRDS("extdata/bristol-ttwa.rds")
 
 The origin and destination zones used in this chapter are the same: officially defined zones of intermediate geographic resolution (their [official](https://www.ons.gov.uk/peoplepopulationandcommunity/populationandmigration/populationestimates/bulletins/annualsmallareapopulationestimates/2014-10-23) name is Middle layer Super Output Areas or MSOAs).
 Each house around 8,000 people.
-Such administrative zones can provide vital context to transport analysis, such as the type of people who benefit particular interventions.
+Such administrative zones can provide vital context to transport analysis, such as the type of people who might benefit most from particular interventions <!-- reference --> .
```
jannes-m: A reference might be helpful.

08-transport.Rmd
```diff
@@ -143,10 +151,11 @@ region_ttwa = readRDS("extdata/bristol-ttwa.rds")
...
 
-The geographic resolution of these zones is important: small zones with high geographic resolution are usually preferable but their high number in large regions can have consequences for processing (especially for origin-destination analysis in which the number of possibilities increases as a non-linear function of the number of zones).
-Another issue with small zones is related to anonymity rules: more socio-demographic variables are often available for large zones.
+The geographic resolution of these zones is important: small zones with high geographic resolution are usually preferable but their high number in large regions can have consequences for processing (especially for origin-destination analysis in which the number of possibilities increases as a non-linear function of the number of zones) <!-- reference -->.
```
jannes-m: a reference might be helpful

08-transport.Rmd
```diff
@@ -143,10 +151,11 @@ region_ttwa = readRDS("extdata/bristol-ttwa.rds")
...
+Another issue with small zones is related to anonymity rules, that means data will usually only be made available in such such a form that it is impossible to infer from them the identity of individuals. 
+In any case, fine-scale socio-demographic variables are often hard to collect or expensive which is why a common approach is to use sociodemographic data from coarser resolutions <!-- such as? And also think of the socio-demographic data we are using in the location chapter, 1 km^2 is already pretty fine and some data is even available for a 100 m resolution -->.
```
jannes-m: Pls check if my rewrite captures the meaning you want to convey. Additionally, what are larger zones or coarser resolutions in this context, i.e., an example?

08-transport.Rmd
```diff
@@ -170,12 +179,12 @@ aggregation of the origin-destination (OD) data to the level of residential zone
...
 od = readRDS("extdata/bristol-od.rds")
 zones_attr = group_by(od, o) %>% 
-  summarise_if(is.numeric, sum) %>% 
+  summarize_if(is.numeric, sum) %>% 
   rename(geo_code = o)
 ```
 
 What just happened?
-After the `od` dataset was read-in, the chained operation:
+After the `od` dataset was read-in<!-- I think a short description of the od dataset might be helpful here -->, the chained operation:
```
jannes-m: I know you say that a dataset description will follow in the next section. I think a description here would be rather helpful already here.

08-transport.Rmd
```diff
@@ -190,8 +199,14 @@ We can verify that the IDs match the zone dataset with the `%in%` operator as fo
 summary(zones_attr$geo_code %in% zones$geo_code)
 ```
 
+<!-- but this way you only check if a is in b but not if b is also in a, i.e., you won't notice if zones has more elements in geo_code than zones_attr, what I always do is:
```
jannes-m: suggestion for id consistency checking

08-transport.Rmd
```diff
@@ -203,21 +218,24 @@ The result is an updated `zones` dataset that contains new columns representing
 The geographic distribution of trip origins is illustrated in the left-hand map in Figure \@ref(fig:zones).
 This shows that most zones have between 0 and 4,000 trips originating from them in the study area.
 More trips are made by people living near the center of Bristol and fewer on the outskirts.
-Why is this? Remember that we are only dealing trips within the study region:
-low trip numbers in the outskirts of the region can be explained by the fact that many people in these peripheral zones will travel to other regions.
+<!-- What is the reasoning for this? I mean since it is an aggregate we could also consider all trips made regardless of the fact if the destiation is within the study area. -->
+Why is this? Remember that we are only dealing with trips within the study region:
```
jannes-m: What is the reasoning for this? I mean since it is an aggregate we could also consider all trips made regardless of the fact if the destiation is within the study area.

08-transport.Rmd
```diff
@@ -203,21 +218,24 @@ The result is an updated `zones` dataset that contains new columns representing
...
 zones = group_by(od, d) %>% 
-  summarise_if(is.numeric, sum) %>% 
-  select(geo_code = d, all_dest = all) %>% 
+  summarize_if(is.numeric, sum) %>% 
+  dplyr::select(geo_code = d, all_dest = all) %>% 
   inner_join(zones, ., by = "geo_code")
 ```
 
+
+<!-- so od contains commute trips? As I have mentioned before it might be a good idea to already explain the dataset in this section. -->
```
jannes-m: Again, od description might be really helpful in this section already

08-transport.Rmd
```diff
@@ -229,24 +247,26 @@ qtm(zones, c("all", "all_dest"), fill.palette = viridis::plasma(5),
...
 ```
 
+<!-- I guess that geocode_1 refers to origin and geocode_2 refers to destination, but in the text you talk of columns o and d, pls change acoordingly. -->
 ```{r od, echo=FALSE}
 od_top5 %>% 
   knitr::kable(caption = "Sample of the origin-destination data stored in the data frame object `od`. These represent the top 5 most common desire lines between zones in the study area.")
 ```
 
-This table provides a snapshot of Bristolian travel patterns.
+<!-- How can you infer intrazonal trips from the table? Do you mean interzonal, i.e., trips between zones? -->
```
jannes-m: How can you infer intrazonal trips from the table? Do you mean interzonal, i.e., trips between zones? Ah, now I see, row 5 refers to an intrazonal trip! Ok, then refer to row 5 in the text, otherwise the reader is forced to search and to find out that there are indeed intrazonal pairs.

08-transport.Rmd
Robinlovelace: Fair point - as a geographer whose main area of application is transport you can understand my exuberance!

08-transport.Rmd
Robinlovelace: It's a term that links all those things together. If you've an alternative suggestion let me know but good to get that in there I think, linking back to the geocomputation lit (before it was called geocomputation when Tobler coined the '1st law').

08-transport.Rmd
Robinlovelace: Good suggestion, will do.

08-transport.Rmd
Robinlovelace: I believe so. Hopefully that will change and I remember having second thoughts about that statement. I plan to tone-it-down (or feel free to tone-down yourself in this PR if you think of an appropriate way of saying 'full agent based transport models running at high temporal resolution are complex compared with aggregate level models' here)

08-transport.Rmd
Robinlovelace: Yep - potentially confusing. Can you think of a less confusing way of phrasing this?

08-transport.Rmd
Robinlovelace: Yes - areas in which people who live there also work there.

08-transport.Rmd
Robinlovelace: OK good point - pls clarify it (the table changed, that may explain some confusion here).

Robinlovelace:
Thanks for the review so far - will wait until you say 'OK' before merging it and acting on your various enlightening comments.

08-transport.Rmd
```diff
@@ -269,16 +289,18 @@ desire_lines = od2line(od_inter, zones)
...
+Next it would be interesting to have a look at the distribution of interzonal modes, e.g., between which zones is cycling the least, the most common means of transport. <!-- maybe an idea for an exercise? -->
 <!-- These include Bradford Stoke to the North and Portishead to the West. -->
 
 ```{r, eval=FALSE}
 plot(desire_lines$geometry, lwd = desire_lines$all / 500)
 ```
 
-```{r desire, echo=FALSE, warning=FALSE, message=FALSE, fig.cap="Desire lines representing trip patterns in the Bristol Travel to Work Area. The 4 black lines represent interzonal OD pairs in Table 8.1."}
+<!-- might be a bit confusing that the four black lines are shown but not created in the code chunk, maybe add clarification -->
+```{r desire, echo=FALSE, warning=FALSE, message=FALSE, fig.cap="Desire lines representing trip patterns in the Bristol Travel to Work Area. The four black lines represent interzonal OD pairs in Table 8.1."}
```
jannes-m: might be a bit confusing that the four black lines are shown but not created in the code chunk, maybe add clarification

08-transport.Rmd
```diff
@@ -300,14 +322,16 @@ Desire lines contain only two vertices (their beginning and end points) but rout
...
-Before proceeding, however, you should be aware of the disadvantages of online routing services: they can be slow due to their reliance on internet connections and expensive.
+Before proceeding, however, you should be aware of the disadvantages of online routing services: they are slow or even inaccessible when accessed with a bad internet connection and can be expensive, i.e., not all services are free of charge.
+For instance, the Google routing API sets the limit to 2500 free queries per day ^[`ggmap::route()` lets you use Google's routing API through R.].
 <!-- Todo: add link to Mark's presentation on dodgr vs Google routing costs (RL) -->
 
-A popular and free (for limited uses) routing service is the Open Source Routing Machine ([OSRM](http://project-osrm.org/)).
-Instead of routing *all* desire lines generated in the previous section, which would be time and memory-consuming, we will focus attention on only the desire lines of policy interest.
+A popular and free (for limited uses) routing service is the Open Source Routing Machine ([OSRM](http://project-osrm.org/)).^[Another popular OSM routing API is https://openrouteservice.org/.]
```
jannes-m: GIScience Heidelberg is developing https://openrouteservice.org and I think it is not related to OSRM, right?

08-transport.Rmd
```diff
@@ -300,26 +322,34 @@ Desire lines contain only two vertices (their beginning and end points) but rout
...
 Based on this reasoning
 <!-- Todo: add references supporting this (RL) -->
-we will only route desire lines along which a high (300+) number of car trips take place that are up to 5 km in distance:
+we will only route desire lines along which a high (300+) number of car trips take place that are up to 5 km in distance.
 
+<!-- st_length works with sf 0.6.0 but not with 0.6.1 -->
 ```{r, message=FALSE}
 desire_lines$distance = as.numeric(st_length(desire_lines))
-desire_carshort = filter(desire_lines, car_driver > 300, distance < 5000)
+desire_carshort = dplyr::filter(desire_lines, car_driver > 300 & distance < 5000)
```
jannes-m: I think this is more readable

08-transport.Rmd
```diff
@@ -300,26 +322,34 @@ Desire lines contain only two vertices (their beginning and end points) but rout
...
+A popular and free (for limited uses) routing service is the Open Source Routing Machine ([OSRM](http://project-osrm.org/)).^[Another popular OSM routing API is https://openrouteservice.org/.]
+Instead of routing *all* desire lines generated in the previous section, which would be time and memory-consuming, we will focus on the desire lines of policy interest.
 The benefits of cycling trips are greatest when they replace car trips.
 Clearly not all car trips can realistically be replaced by cycling, but 5 km Euclidean distance (or around 6-8 km of route distance) is easily accessible within 30 minutes for most people.
 Based on this reasoning
 <!-- Todo: add references supporting this (RL) -->
-we will only route desire lines along which a high (300+) number of car trips take place that are up to 5 km in distance:
+we will only route desire lines along which a high (300+) number of car trips take place that are up to 5 km in distance.
 
+<!-- st_length works with sf 0.6.0 but not with 0.6.1 -->
```
jannes-m: just a remark: on my computer `st_length` worked with sf 0.6.0 but not with 0.6.1

08-transport.Rmd
```diff
@@ -300,26 +322,34 @@ Desire lines contain only two vertices (their beginning and end points) but rout
...
 ```{r, message=FALSE}
 desire_lines$distance = as.numeric(st_length(desire_lines))
-desire_carshort = filter(desire_lines, car_driver > 300, distance < 5000)
+desire_carshort = dplyr::filter(desire_lines, car_driver > 300 & distance < 5000)
 route_carshort = line2route(desire_carshort, route_fun = route_osrm)
 ```
 
+`st_length()` determines the length of a linestring, and falls into the distance relations category (see also section \@ref(distance-relations)).
+<!-- could you pls explain a bit more in detail what kind of input the line2route function expects. In the help it says it expects from and to coordinates. But here you use a simple feature. Since this is a line I am not sure what coordinates the function uses. -->
```
jannes-m: Could you pls explain a bit more in detail what kind of input the line2route function expects. In the function help it says it expects from and to coordinates. But here you use a simple feature. Since this is a line I am not sure what coordinates the function uses.

08-transport.Rmd
```diff
@@ -329,17 +359,22 @@ The following command makes use of the ability of simple features objects to con
 desire_carshort$geom_car = route_carshort$geometry
 ```
 
-This allows for the desire lines along which many short car journeys take place to be plotted alongside likely routes traveled by cars, with the width of the routes proportional to the number of car journeys that could potentially be replaced.
+This allows plotting the desire lines along which many short car journeys take place alongside likely routes traveled by cars, with the width of the routes proportional to the number of car journeys that could potentially be replaced.
 The code below results in Figure \@ref(fig:routes), which demonstrates along which routes people are driving short distances:
 
+<!--
+1. Why is the starting point for both routes different (shouldn't it be the polygon centroid)? Or the centroid is in the middle of nowhere. Then say so in the caption and/or use leaflet in the background.
```
jannes-m: fig:routes remarks
1. Why is the starting point for both routes different (shouldn't it be the polygon centroid)? Or the centroid is in the middle of nowhere. Then say so in the caption and/or use leaflet in the background.
2. Caption: car journeys are made (red); As far as I understood you used polygon centroids as starting points. Then you computed the route between centroids. I wouldn't say that this route is representative or the most used route. It is just the shortest route between two centroids. Since we have only aggregated data on the polygon level, we simply have no clue what the most frequent point origins and destinations are. 
3. Plotting the centroids and/or the polygons would also help visual inspection.

08-transport.Rmd
jannes-m: ok, done that.

08-transport.Rmd
Robinlovelace: Looks like it's not looking briefly at their source code. It's an OSM routing service thought, thanks for flagging it - looks good.

08-transport.Rmd
Robinlovelace: OK yes the documentation of `stplanr::line2route()` needs updating. I have been planning a generic `route()` function that takes a wide variety of inputs (coordinates, lines, text strings of addresses) and creates routes accordingly. Seem like a good idea to do that and include that here?

Here's the beginning of it btw: https://github.com/ropensci/stplanr/blob/master/R/route.R

08-transport.Rmd
```diff
@@ -359,11 +393,12 @@ There are two types of transport nodes:
 2. Nodes that are a part of transport networks, representing individual pathways, intersections between pathways (junctions) and points for entering or exiting a transport network such as bus stops and train stations.
 
 From a mathematical perspective transport networks are simply large graphs.
+<!-- not entirely sure what you mean here: are you talking about connecting a node outside of the network by finding the closest node of the graph network? -->
 The first type of node can be connected to the network with "centroid connectors" [@hollander_transport_2016], the geographic location of which can influence model results, by simulating excessive traffic on streets surrounding zone centroids [@jafari_investigation_2015].
```
jannes-m: Not entirely sure what you mean here: are you talking about connecting a node outside of the network by finding the closest node of the graph network?

08-transport.Rmd
```diff
@@ -426,10 +464,13 @@ mat_rail_o = mat_rail[knn_orig, ]
 mat_rail_d = mat_rail[knn_dest, ]
 ```
 
-The final stage is to convert these matrices into meaningful geographic objects, in this case 'multilinestrings' that capture the fact that each stage is a separate line, but part of the same overall trip which can be represented as a simple feature:
+The final stage is to convert these matrices into meaningful geographic objects, in this case simple feature 'multilinestrings' that capture the fact that each stage is a separate line, but part of the same overall trip:
```
jannes-m: in fact, you are not really using multilinestrings, so I think we should rephrase (see my code example below also)

08-transport.Rmd
```diff
@@ -495,8 +552,10 @@ The output of the previous code chunk shows that `ways_sln` is a composite objec
 These include: the spatial component of the network (named `sl`), the graph component (`g`) and the 'weightfield', the edge variable used for shortest path calculation (by default segment distance).
 `ways_sln` is of class `sfNetwork`, defined by the S4 class system.
 This means that each component can be accessed using the `@` operator, which is used below to extract its graph component and process it using the **igraph** package, before plotting the results in geographic space.
+<!-- This might be a knowledge gap, but I have no idea what the edge betweeness is and what it is used for. Could you add a short explanation? -->
```
jannes-m: This might be a knowledge gap, but I have no idea what the edge betweeness is and what it is used for. Could you add a short explanation?

08-transport.Rmd
```diff
@@ -495,8 +552,10 @@ The output of the previous code chunk shows that `ways_sln` is a composite objec
...
 In the example below the 'edge betweeness' is calculated, with the results demonstrating that each graph edge represents a segment: the segments near the center of the road network have the greatest betweeness scores.
 
+<!-- figure caption is missing -->
 ```{r}
 g = ways_sln@g
```
jannes-m: figure caption is missing

08-transport.Rmd
```diff
@@ -510,9 +569,14 @@ igraph::V(g)$m = m$membership
...
-
 One can also find the shortest route between origins and destinations using this graph representation of the route network:
 
+<!-- 
+1. figure caption is missing 
+2. not really clear what you are doing here, what is sum_network_routes doing, why 1 and 20, what is length referring to?
+3. Since this is the local routing we should back-reference remote routing
+4. Why is there a section routes and a separate section route networks?
+-->
 ```{r}
```
jannes-m: 1. figure caption is missing 
2. not really clear what you are doing here, what is sum_network_routes doing, why 1 and 20, what is length referring to?
3. Since this is the local routing we should back-reference remote routing
4. Why is there a section routes and a separate section route networks?

08-transport.Rmd
```diff
@@ -523,13 +587,14 @@ plot(ways_sln@sl$geometry, add = TRUE)
...
 
-In fact we have already done all the hard work needed to priorities cycling infrastructure:
-we have short but car-dependent commuting routes (generated from desire lines) created in section \@ref(routes), desire lines representing the first leg to public transport nodes created in section \@ref(nodes) and the ability to analyse route networks in terms of 'betweenness' and other conceptions for graph theory, covered in the previous section.
-Rather than describe how each of these elements will be combined the final code chunk should speak for itself:
+In fact we have already done all the hard work needed where to prioritize cycling infrastructure:
+we have short but car-dependent commuting routes (generated from desire lines) created in section \@ref(routes), desire lines representing the first leg to public transport nodes created in section \@ref(nodes) and the ability to analyze route networks in terms of 'betweenness' and other conceptions for graph theory, covered in the previous section.
+<!-- maybe it would be better to explain the code a little bit-->
```
jannes-m: I think the code needs some explanation.

08-transport.Rmd
```diff
@@ -558,31 +623,32 @@ m_leaflet = qtm(ways, lines.col = "highway", lines.lwd = 3, lines.palette = c("l
...
+More importantly, from a learning perspective, it has demonstrated how representing transport systems at multiple interacting geographical levels can lead to new insights about how they work, and how to change them.
+However, there is much more that could be done in this area.
 Geocomputation for transport applications is a small but rapidly growing field and space requirements mean that we have only scratched the surface of what is possible, while emphasizing the principles and key methods of implementation in R.
 
 It is possible to build on the foundations presented in this chapter in many directions.
-Transport is the fastest growing source of greenhouse gas emissions in many countries.
-Because of the highly unequal distribution of transport-related emissions across society, and the fact that transport (unlike food and heating) is not essential for well-being, there is great potential for the sector to rapidly decarbonise through demand reduction, electrification of the vehicle fleet and the uptake of active travel modes such as walking and cycling.
+Transport is the fastest growing source of greenhouse gas emissions in many countries <!-- reference -->.
```
jannes-m: add reference

08-transport.Rmd
```diff
@@ -558,31 +623,32 @@ m_leaflet = qtm(ways, lines.col = "highway", lines.lwd = 3, lines.palette = c("l
 tmap::tmap_leaflet(m_leaflet)
 ```
 
+<!-- Maybe we should mention and/or justify why we think that the route between polygon centroids is probably also the most used one. What if most travellers travel from the northwestern part of polygon A to the northeastern part of polygon B?  -->
```
jannes-m: Maybe we should mention and/or justify why we think that the route between two polygon centroids is probably also the most used one. What if most travellers travel from the northwestern part of polygon A to the northeastern part of polygon B?

08-transport.Rmd
```diff
@@ -594,7 +660,7 @@ One growing area of interest surrounds the simulation of individual people and v
 R has the capability to model people in zones, and interface with ABM software such as NetLogo.
 Combined with its geographical capabilities, demonstrated in this book, R would seem an appropriate language for such developments to take place.
 Of course this should be done in ways that build-on and extend existing work in the area.
-R's 'ecological niche' in transport modelling software could be around the integration of detailed geographic analysis techniques with more widely known and used statistical techniques used in transport research.
+R's 'ecological niche' in transport modelling software could be around the integration of detailed geographic analysis techniques with more widely known and used statistical techniques used in transport research.<!-- and advanced statistical technical so far less commonly applied.-->
```
jannes-m: wouldn't be the statistical techniques so far less commonly used in transportation a further goal?

jannes-m:
@Robinlovelace Great chapter, I enjoyed reading it and have learned a lot about transportation and **stplanr**! I hope my comments are useful. But at least you have a bit time to incorporate them.

08-transport.Rmd
Robinlovelace: Discussed this today with @mem48 - yes definitely needs a brief description - thanks for flagging that. The point is treating route networks as graphs opens-up new analysis possibilities. Comments Malcolm?

Robinlovelace:
Great stuff + many thanks Jannes! Will bosh this this wknd taking advantage of non :sun_with_face: conditions!

08-transport.Rmd
Robinlovelace: I disagree that the borders should have a caption: it's cleary which is the inner city boundary and adding another legend would clutter the map. I agree with the comment about caption length and have reduced it accordingly.

08-transport.Rmd
Robinlovelace: Happy to consider further updates but think this footnote addresses the main points: 

```
In this plot the origins of the red routes and black desire lines are not identical.
This is because zone centroids rarely lie on the route network: instead the route originate from the transport network node nearest the centroid.
Note also that routes are assumed to originate in the zone centroids, a simplifying assumption which is used in transport models to reduce the computational resources needed to calculate the shortest path between all combinations of possible origins and destinations [@hollander_transport_2016].
```

08-transport.Rmd
Robinlovelace: In reply @jannes-m - these are good comments worthy of responses:

1. No longer plotted - was a plot-heavy chapter and can be a reward encouraging readers to run the code.
2. I've explained what's going on and that 1 and 20 are arbitrary (sure a better example could be found but this is good to demonstrate local routing I think) and explained how to select nodes by geographic location.
3. Done
4. To coincide with the principle of structuring our book, and the chapters within it, based on difficulty and frequency with which the methods will be needed (routes are easier and more commonly needed than route networks)

Sound reasonable? Seem fixed to you?

08-transport.Rmd
Robinlovelace: Explained now - agreed I got lazy!

08-transport.Rmd
Robinlovelace: Sorted

08-transport.Rmd
Robinlovelace: Addressed - the new wording advocates this.

08-transport.Rmd
Robinlovelace: Good point - addressed.

Robinlovelace:
Hi @jannes-m many thanks for the detailed review! I think I've acted on or in some cases replied to all your questions - let me know if you've any further questions.

08-transport.Rmd
jannes-m: 👍

08-transport.Rmd
jannes-m: 👍

08-transport.Rmd
jannes-m: ok

08-transport.Rmd
jannes-m: Yes, perfect as is. And thanks for putting up with my comments!

08-transport.Rmd
jannes-m: Yes, perfect! Only one remark with regard to 4. This is a good argument. Maybe we should state so in the text or at the beginning of route networks, that they are more complex and therefore presented in a separate section

08-transport.Rmd
jannes-m: 😄  but I perfectly understand!

08-transport.Rmd
Robinlovelace: No problem, thanks for taking the time to make detailed comments, really useful internal review that will make the external review stronger.

PR #145, opened 2018-01-25
Add coffee map example

Robinlovelace:


Robinlovelace:
Generates this map: 
image

jannes-m:
👍  I really like coffee! I am not entirely sure about what the legend tells us, sth. like tons or bags of coffee produced/exported? If this is the case, it would be more just to refer this to the number of people or country area (or even better farming area per country...) or both, e.g., kg of coffee/km^2 or kg or coffee/inhabitants or even both kg /km^2/inhabitants. Well, depends on what one wants to show, if it's absolute terms, then we should explicitly state this. I would expect the map would change dramatically when accounting for area or people. Vietnam would be super-super brown, and Brazil maybe should light-yellow.

Robinlovelace:
The legend is fixed in the latest version. I'm now thinking a) of using this as the main join example in c3 and b) adding the minimal `coffee_data` data object to **spData** (it's tiny). Sound like a plan @Nowosad ?

Nowosad:
I find it impossible to say no to coffee;)

jannes-m:
I totally agree, I will grab just another cup of ☕️ !

PR #156, opened 2018-02-05
Rename chapters, add placeholders

Robinlovelace:
Following-on from our conversation and considering 'content is king' and that we can always revisit this running-order, please @jannes-m and @Nowosad take a look at this proposal and let me know. I've changed the order of a few things. I think it makes sense to teach people how to use other peoples' algorithms (in the GIS chapter) before teaching them how to create their own ones. I think the solution (renewable energy) should come after the problem (SLR). Anything else I've missed? Anything you'd like to see changed?


14-eco.Rmd
```diff
@@ -1,4 +1,6 @@
-# Ecological case study
+# (PART) Geocomputation in the wild {-}
```
Robinlovelace: Note the more interesting part title. More exciting than Part I and II. The distinction is that these latter chapters are more 'wild' in terms of difficulty level is the thinking.

jannes-m: Good thinking!

jannes-m:
Sounds good, I think the structure is developing.
Not sure how you named Part 3 and 4 in this PR.
And I think when describing the structure of the book at the beginning of the book and at the beginning of Part 3, we should point out, that while we have laid the foundations (pretty straight-forward what one has to learn to do geocomputation in R) in the beginning in a comprehensive manner, the third part is more a collection of advanced geocomputational methods and by no means comprehensive. Here, we try to provide contents which have not been introduced in other textbooks. However, we will point to these textbooks where appropriate (e.g., kriging interpolation).

Robinlovelace:
Hi @jannes-m I've thought long about part names and tried various combinations. Reasoning behind current suggestions:

- Benefits of stating that some were 'methods'
- `Basic geocomputing methods` is rather long and `Basic methods` vs `Advanced methods` is accurate I think
- In action vs in wild still there - but could change

PR #159, opened 2018-02-07
Update 11-gis.Rmd

gisma:
The updated `link2GI` package is is on the way to CRAN
The old call was not supported since version 0.2.0. I tried to simplify this kind of usage adding the wrapper function `findGRASS` and for convenient usage of `linkGRASS7` I have added a EPSG4326 worldwide location by default. This makes it much more comfortable to use just `linkGRASS7` and adapt the setting according to your needs.


11-gis.Rmd
```diff
@@ -328,11 +328,7 @@ First of all, we need to find out if and where GRASS7 is installed on the comput
 
 ```{r}
 library(link2GI)
-if (Sys.info()["sysname"] == "Windows") {
-  link = link2GI::searchGRASSW()
```
Robinlovelace: Nice - more OS-agnostic.

jannes-m:
@gisma To make it less of a hazard, I thought I just copied your code manually. However, your code still throws an error on my Windows machine. I think we have encountered this problem before, and back then you told me that the problem was that OSGeo4W installs GRASS6 and GRASS7.

```r
library(link2GI)
link = findGRASS() 
link2GI::linkGRASS7()

Error in rgrass7::initGRASS(gisBase = grass$gisbase_GRASS, home = tmpDir(),  : 
  
### rgrass7 is not compatible with the GRASS GIS version 'GRASS 6.4.3 (2013) '!
### Please use the package appropriate to the GRASS GIS version:
### GRASS GIS Version 5.x.y  --  GRASS
### GRASS GIS Version 6.x.y  --  spgrass6
### GRASS GIS Version 7.x.y  --  rgrass7
```

gisma:
@jannes-m  Ah ok, if I have already noticed it then probably I have forgotten it. Well this is not nice at all. When I started with `link2GI` I made the decision that it is far out interest to support the old versions of GRASS. And I am still convinced that there is in alldays reality no need to do so. To make it short, I think it would be better to remove `link2GI` from the book chapter because I have right in the moment just no time to develop this for all old GRASS  versions under (old) Windows systems.  However I am open for any  Suggestions...

jannes-m:
I think it is not about supporting GRASS6 but more about making linkGRASS7 recognize the path to a GRASS7 installation. This is basically, what I was doing in the code you were editing:

```r
library(link2GI)
library(rgrass7)
link = findGRASS() 
# find a GRASS7 installation, and use the first one
ind = grep("7", link$version)[1]
# next line of code only necessary if we want to use GRASS as installed by 
# OSGeo4W. Among others, this adds some paths to PATH, which are also needed
# for running GRASS.
link2GI::paramGRASSw(link[ind, ])
grass_path = ifelse(!is.null(link$installation_type) && 
                      link$installation_type[ind] == "osgeo4W",
                    file.path(link$instDir[ind], "apps/grass", link$version[ind]),
                    link$instDir)
initGRASS(gisBase = grass_path,
          # home parameter necessary under UNIX-based systems
          home = tempdir(),
          gisDbase = tempdir(), location = "london", 
          mapset = "PERMANENT", override = TRUE)
```

On the one hand, I think this might be useful to implement this since many users will have installed GRASS6 & 7 via OSGEO4W. On the other hand, maybe it is not worth the effort since I believe QGIS 3 is only supporting GRASS7 (see also http://lists.osgeo.org/pipermail/qgis-user/2016-November/038539.html) and will become the QGIS LTR next February. Let me know what you think.

PR #161, opened 2018-02-08
Shift spatial cv chapter up

Robinlovelace:
@jannes-m you happy with this? Thinking: more consistent and should build on the raster-vector data.


jannes-m:
I am not entirely sure what you are planning to write in the algorithm chapter. In the end we should probably order the chapters by degree of complexity (and maybe by fit, i.e. some chapters might naturally belong together). So I would suggest to decide this when the chapters are finished.

Robinlovelace:
Sounds reasonable.

PR #163, opened 2018-02-13
Hide part 4

Robinlovelace:
I was initially planning to remove all in-development chapters except c9 but that seemed too drastic. On the other hand keeping them all could lead to confusion and false expectations about current content. As a compromise I suggest we build all chapters in part 3 but hide part 4 chapters. Committing this as a PR to check @Nowosad and @jannes-m are happy with this. Please approve if so!


Nowosad:
Fine with me

jannes-m:
Sounds good!

PR #169, opened 2018-03-01
 adapt link2GI function calls to link2GI >=0.3-1

gisma:
hi all 
I have adapted and reduced  the link2GI function calls to the (upcoming) CRAN  versions >=  0.3-1.
It is also working with the new RSAGA 1.00 even if this package is performing an (hidden) seperate search.  Please check and if you like integrate it. 
cheers Chris


Robinlovelace:
You happy with this @jannes-m ?

jannes-m:
I haven't tried but I will do so when I review the GIS chapter again. For now, we can merge I'd say. And thanks so much for your contributions @gisma!

PR #189, opened 2018-04-06
chapter 13 : adding Docker as fourth solution for parallelized computing under Windows

pokyah:
Hi !
Useful book to me as I'm beginning to build a weather app providing spatialized weather data under R ! 
 
Docker might also be an efficient solution for Windows users as it allows to quickly configure and run an Unix based-OS with a precise set of software version and dependencies. If you want to know more, you can check my [tutorial](https://pokyah.github.io/howto/using-r-with-docker/) about using R with docker .

Thanks for reviewing

### DIFF
--- a/13-spatial-cv.Rmd
+++ b/13-spatial-cv.Rmd
@@ -533,12 +533,13 @@ So before starting the actual resampling it would be wise to reduce runtime with
 In general, multicore processing is easier on Unix-based systems than on Windows systems.
 In fact, cloud computing is usually done and developed on Linux servers.
 Therefore, we will present how to do nested cross-validation using a parallelization approach working only under Unix-based systems.^[Note also that the `mc.set.seeds` parameter used later on is only available on Unix-based systems].
-Windows users have at least three options:
+Windows users have at least four options:
 
 1. Run the resampling without parallelizing it though this will take most likely more than half a day.
 2. Install a virtual machine (e.g. [Oracle VirtualBox](https://www.virtualbox.org/)) to reproduce the parallelization code.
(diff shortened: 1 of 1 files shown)


Robinlovelace:
This looks good to me, thanks for the input @pokyah. Will leave it for @Nowosad and @jannes-m and comment / merge.

jannes-m:
Yes, you are right, this is another viable solution!

pokyah:
You are welcome !

znmeb:
A couple of cautions on Docker for Windows:

1. There are two versions of Docker that run on Windows - Docker for Windows and the older Docker Toolbox. Docker for Windows requires Windows 10 Pro - no other version of Windows is supported. If you have any other version of Windows, you need Docker Toolbox.
2. In either case, Docker operates inside a full virtual machine running Linux. Thus it has available at most half of the RAM in your desktop / laptop, and often less than half. For example, my Windows 10 Pro laptop has 8 GB of RAM, but by default Docker containers only have 2 GB to work with. I've been able to push it up to 3 GB but not to 4.

Because of the virtual machine. there is also a measurable processor overhead associated with Docker on a Windows host. I don't know what the exact tradeoffs are, but I recommend running native tools on Windows whenever possible.

jannes-m:
@znmeb You are perfectly right. In terms of parallelization it does not make too much sense but at least you could run the code. In any case, I will probably rewrite the corresponding paragraph by providing also an example how to parallelize the code on a Windows machine (but then without `mc.set.seed`).

znmeb:
@jannes-m Isn't there a parallel random number generator on Windows? It's been a long time since I used Windows R.

PR #194, opened 2018-04-14
Rev map jm

jannes-m:
Hey there,
your chapter starts to look great. Some refining, polishing, restructuring and writing is probably still needed.


09-mapping.Rmd
```diff
@@ -52,11 +53,13 @@ They are fixed images that can be included in printed outputs or published onlin
...
 <!-- well, base R plotting and the grid package already allow for the making of high-quality maps suitable for publication. So the advantage of tmap is to do the same in with less and concise code? In any case, it would be good to emphasize the advantage the specific advantage of tmap compared to base plotting + grid. It would be also worth noting why tmap is the better option than ggplot2. Allows a wide variety of spatial formats -->
 The focus of this section, however, is map making publication-quality and engaging maps using minimal time and code with dedicated packages, particularly **tmap**.
+<!-- well, base R plotting and the grid package already allow for the making of high-quality maps suitable for publication. So the advantage of tmap is to do the same in with less and concise code? In any case, it would be good to emphasize the advantage the specific advantage of tmap compared to base plotting + grid. It would be also worth noting why tmap is the better option than ggplot2.
```
jannes-m: !-- well, base R plotting and the grid package already allow for the making of high-quality maps suitable for publication. So the advantage of tmap is to do the same in with less and concise code? In any case, it would be good to emphasize the advantage the specific advantage of tmap compared to base plotting + grid. It would be also worth noting why tmap is the better option than ggplot2.
+Robin: tmap allows a wide variety of spatial formats especially raster objects (probably will also support stars?)

09-mapping.Rmd
```diff
@@ -122,6 +136,7 @@ map_nz1 = map_nz +
 ```
 
 Building on the previously created `map_nz` object the preceding code creates a new map object `map_nz1` which contains another shape (`nz_height`), representing high points (see Figure \@ref(fig:tmlayers), left).
+<!-- tm_bubbles explanation is missing -->
```
jannes-m: tm_bubbles explanation is missing

09-mapping.Rmd
```diff
@@ -182,8 +197,10 @@ tm_shape(nz) + tm_fill(col = 1:nrow(nz)) # fails:
 ```
 
 Instead `col` (and other aesthetics that can vary such as `lwd` for line layers and `size` for point layers) requires a character string naming an attribute associated with the geometry to be plotted.
-Thus one would achieve the desired (plotted in the right-hand panel of Figure \@ref(fig:tmcol)) result as follows:^[
-Figure \@ref(fig:tmcol) also demonstrates another benefit of **tmap**: it automatically converts numeric variables into sensible, graduated bins.
+Thus, one would achieve the desired (plotted in the right-hand panel of Figure \@ref(fig:tmcol)) result as follows:^[
+<!-- wouldn't say sensible, since here it appears as if tmap simply uses equal class sizes...-->
+Figure \@ref(fig:tmcol) also demonstrates another benefit of **tmap**:
```
jannes-m: wouldn't say sensible, since here it appears as if tmap simply uses equal class sizes...

09-mapping.Rmd
```diff
@@ -227,23 +244,26 @@ tm_shape(nz) + tm_fill(col = "AREA_SQ_KM", palette = "RdBu")
 source("code/09-tmpal.R", print.eval = TRUE)
 ```
 
+<!-- sth. wrong with the next sentence -->
 A variable, which name is provided in the `col` argument, is represented by a color palette.
```
jannes-m: Sth. wrong with this sentence.

09-mapping.Rmd
```diff
@@ -277,9 +297,11 @@ many_palette_plotter(c("PRGn", "YlGn", "Set2"), 7,
 <!-- how to reverse palettes with - -->
 <!-- references -->
 
+<!-- This is quite an abrupt change of topics and does not connect well with the previous paragraphs on palette options. -->
 The **tmap** package offers many methods for specifying how variables are plotted on the map, in addition to setting `breaks` manually as we have already seen in this subsection.
```
jannes-m: This is quite an abrupt change of topics and does not connect well with the previous paragraphs on palette options.

09-mapping.Rmd
```diff
@@ -277,9 +297,11 @@ many_palette_plotter(c("PRGn", "YlGn", "Set2"), 7,
...
 They are set with the `style` argument and some of the most useful methods are illustrated in Figure \@ref(fig:break-styles).
 
+<!-- Is there a way to improve the legends, so far they are not really appealaing -->
 ```{r break-styles, fig.cap="Illustration of different binning methods set using the style argument in tmap.", echo=FALSE}
```
jannes-m: Is there a way to improve the legends, so far they are not really appealing.

09-mapping.Rmd
```diff
@@ -291,21 +313,24 @@ The `equal` style divides the range of input values into subranges of equal rang
...
 On the other hand, it could produce categories with a large number of elements (points, areas) and those without any observation.
 To have the same number of elements in each category, one can use the `quantile` style. 
-It is useful for ordinal data <!--examples--> and never produce empty classes. 
+It is useful for ordinal data <!--examples--> and never produces empty classes. 
 This style, however, could have very different values in the same category.
 Natural breaks method, also called `jenks`, tries to identify groups of similar values in the data.
 Breaks between classes are created to maximize the differences between categories. 
-This method provides classes that reflect the values distribution, but in the same creates classes that could be meaningful only for this dataset.
+<!-- what do you mean by could be meaningful only for this? Also "Importantly" sentence sounds strange -->
+This method provides classes that reflect the values distribution, but at the same time creates classes that could be meaningful only for this dataset.
```
jannes-m: what do you mean by could be meaningful only for this?

09-mapping.Rmd
```diff
@@ -291,21 +313,24 @@ The `equal` style divides the range of input values into subranges of equal rang
...
 Importantly, categorical legends described above could be not the most suitable when presenting a large number of objects (e.g. points or polygons) or continuous fields (e.g. continuous rasters).
```
jannes-m: Sentence sounds strange

09-mapping.Rmd
```diff
@@ -291,21 +313,24 @@ The `equal` style divides the range of input values into subranges of equal rang
...
+It assures that each category receives a unique color.
 <!-- references for more info -->
 
 ### Layouts
 
-Layout refers not to the map itself but to its wider surroundings.
+The map layout refers to the combination of all map elements into a cohesive map.
+Map elements include among others the objects to be mapped, the title, the scale bar, margins and aspect ratios.
+<!-- what is the relation of the next two sentences to the general map layout. It sounds more like these should be moved to the palette and break sections. -->
 Color settings relate to the palette and break-points used affect how the map looks.
```
jannes-m: What is the relation of these two sentences to the general map layout? It sounds more like these should be moved to the palette and break sections.

09-mapping.Rmd
```diff
@@ -337,12 +362,13 @@ Some useful layout settings are listed below (see Figure \@ref(fig:layout2) for
...
 
-The impact of changing the color settings listed above is illustrated in Figure \@ref(fig:layout3) (see `?tm_layout` for full list).
+The impact of changing the color settings listed above is illustrated in Figure \@ref(fig:layout3) (see `?tm_layout` for a full list).
 
 ```{r layout3, fig.cap="Illustration of selected color-related layout options.", echo=FALSE}
 source("code/09-layout3.R")
 ```
 
+<!-- Maybe reverse order and progress from the high- to the low-level control. Overall, this section reads a bit like a vignette. Is this the intention. -->
 Beyond the low-level control over layouts and colors, **tmap** also offers high-level styles, using `tm_style_` functions (representing the second meaning of 'style' in the package).
```
jannes-m: Maybe reverse order and progress from the high- to the low-level control. Overall, this section reads a bit like a vignette. Is this the intention?

09-mapping.Rmd
```diff
@@ -399,20 +425,26 @@ The preceding code chunk demonstrates key features of faceted maps created with
...
-The role of spatial visualizations is to provide an information in the most accessible form. 
-This is often done with the use of inset maps.
+An inset map is a smaller map rendered within or next to the main map. 
+<!-- I would say inset maps are usually used for the opposite. This is we have a specific area, say a mountain somewhere in Peru, and the inset map shows were in Peru the mountain is located, see Figure 3 in https://journal.r-project.org/archive/2017/RJ-2017-067/RJ-2017-067.pdf
+and
+http://wiki.gis.com/wiki/index.php/Inset_Map
+So maybe it would be an idea to focus on the NZ Alps raster map as the main map and use NZ within the inset map.
+Maybe this section is also a bit too long. How about shortening it, and maybe use the US example as a challenge in the exercise section.
+-->
+Inset maps usually cover an area with densely located phenomena that cannot be clearly visible at the original map scale.
```
jannes-m: I would say inset maps are usually used for the opposite. This is we have a specific area, say a mountain somewhere in Peru, and the inset map shows were in Peru the mountain is located, see Figure 3 in https://journal.r-project.org/archive/2017/RJ-2017-067/RJ-2017-067.pdf
and
http://wiki.gis.com/wiki/index.php/Inset_Map

So maybe it would be an idea to focus on the NZ Alps raster map as the main map and use NZ within the inset map.

Secondly, this section maybe is also a bit too long. How about shortening it, and maybe use the US example as a challenge in the exercise section.

09-mapping.Rmd
```diff
@@ -542,6 +573,8 @@ The following command creates the animation illustrated in Figure \@ref(fig:urba
 tmap_animation(urb_anim, filename = "urb_anim.gif", delay = 25)
 ```
 
+<!-- Though the second animation is great, I wonder if it is necessary here, especially if the underlying code is not explained in more detail. So this could be also a great challenge for the exercise section?
+-->
 Another illustration of the power of animated maps is provided in Figure \@ref(fig:animus).
```
jannes-m: Though the second animation is great, I wonder if it is necessary here, especially if the underlying code is not explained in more detail and considering the length the viz chapter already has (which even more content to come). So this could be also a great challenge for the exercise section?

09-mapping.Rmd
```diff
@@ -687,11 +720,14 @@ library(globe)
 ```
 
 **tmap** provides a powerful interface for creating a wide range of static maps (section \@ref(interactive-maps) shows how **tmap** also supports interactive maps).
+<!-- if this section is on static map production, it should be part of section static maps -->
 But there are many other options for creating static maps.
```
jannes-m: If this section is on static map production, it should be part of section static maps.

09-mapping.Rmd
```diff
@@ -687,11 +720,14 @@ library(globe)
...
 The aim of this section is to provide a taster of some of these and pointers for additional resources: map making is a surprisingly active area of R package development so there is more to learn than can be covered here.
 
+<!-- maybe this should be separated from basic-map and basic-map-raster
+Also really doesn't belong here since the section header is other mapping packages-->
 The most mature option is to use `plot()` methods provided by core spatial packages **sf** and **raster**, covered in sections \@ref(basic-map) and \@ref(basic-map-raster) respectively.
```
jannes-m: Maybe this should be separated from basic-map and basic-map-raster.
This doesn't belong here since the section header is "other mapping packages".

09-mapping.Rmd
```diff
@@ -727,7 +763,7 @@ g1
 ```
 
 An advantage of **ggplot2** is that it has a strong user-community and many add-on packages.
-Good additional resources can be found in the [ggplot2-book](https://github.com/hadley/ggplot2-book) open source book on the subject, plus descriptions of the multitude of '**gg**packages' such as **ggrepel** and **tidygraph**.
+Good additional resources can be found in the open source [ggplot2-book](https://github.com/hadley/ggplot2-book) and in the descriptions of the multitude of '**gg**packages' such as **ggrepel** and **tidygraph**.
```
jannes-m: is it really open source?

09-mapping.Rmd
```diff
@@ -739,8 +775,8 @@ plotly::ggplotly(g1)
 We have covered mapping with **sf**, **raster** and **ggplot2** packages first because these packages are highly flexible, allowing for the creation of a wide range of static maps.
 Many other static mapping packages are more specific.
 
-Before we cover mapping packages for plotting a specific type of map (in the next paragraph), it is worth considering alternatives to the packages already covered for general-purpose static mapping.
-Table \@ref(tab:map-pkgs) illustrates metrics associated with some commonly used mapping packages, identified using the website [r-pkg.org](https://www.r-pkg.org/search.html?q=map) and accessed via **packagemetrics**.
+<!-- JM: alternative mapping solutions are still missing -->
+Before we cover mapping packages for plotting a specific type of map (in the next paragraph), it is worth considering alternatives to the packages already covered for general-purpose static mapping (Table \@ref(tab:map-pkgs)).
```
jannes-m: Alternative mapping solutions are still missing, just a reminder.

09-mapping.Rmd
```diff
@@ -754,7 +790,7 @@ Table \@ref(tab:map-pkgs) shows a range of mapping packages are available, and t
 Of note is **cartography**, which generates a range of unusual maps including choropleth, 'proportional symbol' and 'flow' maps, each of which is documented in the vignette [`cartography`](https://cran.r-project.org/web/packages/cartography/vignettes/cartography.html).
 
 ## Exercises
-
+<!-- Exercises are still missing -->
```
jannes-m: Exercises are still missing, just a reminder.

Robinlovelace:
Thanks for the review, will merge this then respond to all the comments.

09-mapping.Rmd
Nowosad: :+1:

09-mapping.Rmd
Nowosad: @Robinlovelace

09-mapping.Rmd
```diff
@@ -109,9 +123,9 @@ map_nz = tm_shape(nz) + tm_polygons()
 class(map_nz)
 ```
 
-`map_nz` can be plotted later, for example by adding additional layers (as we'll below) or simply running `map_nz` in the console, which is equivalent to `print(map_nz)`.
+`map_nz` can be plotted later, for example by adding additional layers (as shown below) or simply running `map_nz` in the console, which is equivalent to `print(map_nz)`.
```
Robinlovelace: Nice remove of unnecessary use of the [plural possessive pronoun](https://www.thefreedictionary.com/We+(word)), thanks for this @Jannes-m, I've been caught red-handed (check the blame)!

09-mapping.Rmd
Robinlovelace: Well-spotted. Sorted now, even though `tm_bubbles()` is no longer used in the good. Definite good to mention though so thanks for pointing this out.

09-mapping.Rmd
```diff
@@ -147,14 +162,14 @@ This is demonstrated in the code chunk below which plots `map_nz1` to `map_nz3`,
 tmap_arrange(map_nz1, map_nz2, map_nz3)
 ```
 
-Additional elements such as north arrows, scale bars and layout options can also be added with the `+` operator.
+Additional elements such as north arrows (`tm_compass()`), scale bars (`tm_scale_bar()`) and layout options (`tm_layout()`) can also be added with the `+` operator.
```
Robinlovelace: Great point, thanks for adding this.

09-mapping.Rmd
Robinlovelace: Please take a look at the updated figure here http://geocompr.robinlovelace.net/adv-map.html#fig:break-styles  following this commit which I think fixes this issue: https://github.com/Robinlovelace/geocompr/commit/b3aeda6ecc6be741591c1ef507c8d0ebe894ac92

09-mapping.Rmd
Robinlovelace: Sorted.

09-mapping.Rmd
Robinlovelace: Fair point - could you try to make the changes needed to make this point in the example + prose?

09-mapping.Rmd
Robinlovelace: I disagree: I think we should show people what is possible, encourage them to reproduce by linking to the code and inspire them to change the world ;)

Robinlovelace: I disagree: I think we should show people what is possible, encourage them to reproduce by linking to the code and inspire them to change the world ;)

09-mapping.Rmd
Robinlovelace: No, this provides context that (hopefully) makes the rest of the section make sense. I think `plot()` is justified here because when we say 'other' we mean 'other than **tmap**. **sf** and **raster** are legitimate packages for plotting and deserve mention early in this section (also for context) although this is not their focus. Although happy to see alternatives proposed either as a commit directly to master or a PR (preferably).

09-mapping.Rmd
Robinlovelace: Yes - I linked to the source code (of version II of the book, version I is not).

09-mapping.Rmd
Robinlovelace: True, a priority for the chapter now it's (gradually) coming together. Ideas welcome.

PR #196, opened 2018-04-16
Issue with subsetting and creating layered map plot

erstearns:
There appears to be an issue with the subsetting of the 'world' dataset to Asia-only (lines 252-261). When trying to create the layered plot of Asia atop the world baselayer ["pop], Asia is already highlighted in yellow, which is the plot you show, and then when you 'add' the Asia object, and specify it to be red, it is added, however it is added in the lower right hand corner of the plot, suggesting it has lost it's spatial orientation linked to the original world object. Happy to send a snip if helpful. Not sure how to fix at the moment.


erstearns:
This should have been reported as an 'Issue' - have reported, closing this now.

PR #200, opened 2018-04-17
Spatial cv rev ii rl

Robinlovelace:
Heads-up @jannes-m and @Nowosad think these tweaks are pretty non-controversial but important in pointing people to the established fields of geostatistics, spatial statistics and point patter analysis and some leading resources by leading researchers that provide context on this chapter's contents.

### DIFF
--- a/13-spatial-cv.Rmd
+++ b/13-spatial-cv.Rmd
@@ -23,6 +23,10 @@ Required data will be downloaded in due course.
 
 Statistical learning is concerned with the use of statistical and computational models for identifying patterns in data and predicting from these patterns.
 Because of its original as a statistical programming language, statistical learning is one of R's great strengths R (see section \@ref(software-for-geocomputation)).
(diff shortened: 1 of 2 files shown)


Robinlovelace:
Nudge @jannes-m - seem OK?

jannes-m:
Well, these books are great reference but focus mainly on spatial point patterns, and therefore only somewhat related to the contents of this chapter. I think I would not state them that prominently in the introduction since this might be a bit misleading, hence I'll move them a bit. Thanks for the input!

Robinlovelace:
Only the final book by Badelley is focussed on point patterns. The other 2 are broader, as shown in their ToCs:

- Gelfand et al. (2010) - Part IV is on point patterns: https://books.google.co.uk/books?id=Xf4leslPDzsC&lpg=PA3&ots=mX-Z0fxbrP&dq=Handbook%20of%20Spatial%20Statistics%20table%20of%20contents&pg=PR5#v=onepage&q=Handbook%20of%20Spatial%20Statistics%20table%20of%20contents&f=false

- Diggle et al (2007) seems to have some good stuff on spatial prediction and provides good background on this chapter + has Bayesian chapter although I agree that, like this chapter, it's focussed on point data: https://books.google.co.uk/books/about/Model_based_Geostatistics.html?id=qCqOm39OuFUC&printsec=frontcover&source=kp_read_button&redir_esc=y#v=onepage&q&f=false

PR #201, opened 2018-04-18
Typo: Reference sfc but have sfg written

erstearns:
I believe there is a typo -- you refer to an sfc object but have sfg written, based on my understanding thus far, sfc is the geometry column comprised of sfg objects.


Nowosad:
Thanks!

PR #205, opened 2018-04-19
Referred to incorrect object

erstearns:
In line 527, tried to create a sfg polygon from a list created in section 2.1.5.2, but believe you meant to create from the polygon list immediately preceding the transformation. Just left off a 1.


Nowosad:
Thanks!

PR #210, opened 2018-04-25
Reproducibility: show install.packages code

Robinlovelace:
Aimed at new users not used to installing packages. Also allows the book to be reproduced if spDataLarge cannot be installed - e.g. on managed computers.

Heads-up @jannes-m and @Nowosad happy with this? This is for a situation I'm teaching where users cannot install packages from outside CRAN. Also new users are often confused when 'library(sf)` fails. I think adding this explicit install packages code is a good plan in this 1st practical chapter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ cycle.*
 tmap_style_previews
 urb_anim.gif
 09-us_pop.gif
+spDataLarge*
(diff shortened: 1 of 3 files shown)


jannes-m:
👍 We probably should reference to this at the beginning of each chapter when we are attaching packages?

Robinlovelace:
Yes probably - more concise that showing

```r
install.packages("packagename")
```

At the beginning of every chapter :grinning: I'll assume it's all good an merge this.

PR #217, opened 2018-05-02
Fix typo in tmap annotation

wdearden:
The annotation for the first tmap plots say tmap_shape when the function call is tm_shape.


Robinlovelace:
Great timing: you got in just before the weekly rebuild of the README and are now listed as a contributor: https://github.com/Robinlovelace/geocompr#contributing Many thanks!

Issue #221, opened 2018-05-03
Review 2: responses


Robinlovelace:
- [x] 1.1 I would split chapter 6 into input versus output (visualization). The former, I would add to chapter 2, while the latter could be a little expanded by some of the material taught at Use R! conferences. Yes, I see that there is chapter 9 but data output is not limited to map making. Chapter 2 should then be split into two. 2a would be just about vector and raster data, and 2b about coordinate reference systems, units, data input.
    **- Thank you for this suggestion. We decided to reorganize the chapters, however in a different way than suggested. For example, chapter 5 will be split into two separate chapters: one about geometric operations and another about CRS**
- [x] 1.2 [RL[ Unfortunately, the same (lucid and easy to read) cannot be said about the examples in part II. For example, the osmdata::getbb command on page 155 specifies an output format that is not known to the osmdata package, which in its latest version (as of Feb 2018) allows only for data.frame, matrix, string, or polygon. [RL]
    **- `osmdata::getbb()` was in the development version of **osmdata**. Was fixed on 2018-02-22 when the osmdata 0.0.6 was sent to CRAN - see https://cran.r-project.org/package=osmdata**
   **- We agree that the description of the code in c7 was clunky at times. To address we have re-written the description of the code at the outset of c7 to make it clearer (see this [commit](https://github.com/Robinlovelace/geocompr/pull/225/commits/58f8841395585fa33ae000587fd283bfba38cc6a )) [RL]**
- [x] 1.3 After two Human geography examples, I am missing some physical geography in Part II. Please touch base with Edzer Pebesma and his group to see whether they would share an example or two. [RL, JN, JM - to discuss]
    **- Our next applied chapter will contain examples from physical geography.**
- [x] 1.4 [RL, JN, JM - to discuss] I would leave these sections (5.2) where they are. The topic is important in and by itself and easier to find (even for lookup purposes if the book is used as a reference) if we leave it in section 5.2.
    **- Thank you for your comment.**
- [x] 1.5 [RL, JN, JM - to discuss] (chapter 2) No, please leave it as is. Similar to ASDAR, I believe that it is pedagogically much easier to explain how GIS data is structured by looking at the code. Which is why I have been using ASDAR and would now use your book even in an Intro to GIS class.
    **- Thank you for your comment.**
[...]

PR #225, opened 2018-05-07
Responses to Review 2

Robinlovelace:
Plan: merge this when all issues listed here #221 are addressed in some way. Great start on this.

### DIFF
--- a/08-location.Rmd
+++ b/08-location.Rmd
@@ -494,7 +494,9 @@ For instance, a score greater than 9 might be a suitable threshold indicating ra
 library(leaflet)
 library(sp)
 # have a look at suitable bike shop locations in Berlin
-berlin = metros[metro_names == "Berlin"]
+# berlin = metros[metro_names == "Berlin", ]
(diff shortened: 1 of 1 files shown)


05-transform.Rmd
```diff
@@ -955,7 +955,11 @@ plot(polyg, col = "grey", main = "POLYGON")
 par(par_old)
 ```
 
-This process can be also reversed using `st_cast`:
+Transition from mulitpoint to linestring is a common operation used to convert a set of ordered point observations, such as GPS measurements or geotagged media, into a path.
```
Robinlovelace: I think this line should be:

```
Conversion from mulitpoint to linestring is a common operation used to convert a set of ordered point observations, such as GPS measurements or geotagged media, into a path.
```

Transition implies a gradual process with intermediary steps, as in the **transformr** package (which we should mention!) for example: https://twitter.com/thomasp85/status/982223090561179648

Nowosad: Feel free to change it;)

PR #239, opened 2018-05-15
tiny: noob friendly clarification & referencing

layik:


Robinlovelace:
Gone stale so closing 4 now - thanks for the input and agree with the idea :+1:

PR #248, opened 2018-05-29
Fixed filtering to empty data frame

maxpohlman:
This example would crash when filteredData() returned an empty sf object/data frame (reproducible by setting the threshold to 3300 - 4000). Switching from base-r index slicing to dplyr's filter fixed the problem.

### DIFF
--- a/coffeeApp/app.R
+++ b/coffeeApp/app.R
@@ -1,4 +1,4 @@
-# Credit: build on the exmaple in https://rstudio.github.io/leaflet/shiny.html
+# Credit: build on the example in https://rstudio.github.io/leaflet/shiny.html
 library(sf)
 library(shiny)
 library(spData)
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Thanks for the input. I can reproduce the issue by defining the UI as follows:

```r
ui = bootstrapPage(
  tags$style(type = "text/css", "html, body {width:100%;height:100%}"),
  leafletOutput("map", width = "100%", height = "100%"),
  absolutePanel(top = 10, right = 10,
                sliderInput("range", "Coffee Production", 3300, 4000,
                            value = c(3300, 4000), step = 100
                ),
                selectInput("year", "Year", c(2016, 2017)
                ),
                checkboxInput("legend", "Show legend", FALSE)
  )
)
```

Which generates the following stack trace:

```
Warning: Error in to_ring.default: Don't know how to get polygon data from object of class XY,GEOMETRYCOLLECTION,sfg
Stack trace (innermost first):
    81: to_ring.default
    80: to_ring
    79: to_polygon.default
    78: to_polygon
    77: to_multipolygon.default
    76: FUN
    75: lapply
```

Robinlovelace:
And I confirm this issue is fixed with your commit, a definite improvement: 

image

Robinlovelace:
There is a bigger problem with the app: the map hides the slider! Any ideas how to fix that appreciated...

PR #263, opened 2018-07-29
Update KG

katygregg:
:)

### DIFF
--- a/09-gis.Rmd
+++ b/09-gis.Rmd
@@ -47,22 +47,22 @@ The advantages of a good CLI such as that provided by R are numerous.
 Among others, a CLI:
 
 - Facilitates the automation of repetitive tasks. 
-- Ensures transparency and reproducibility (which also is the backbone of good scientific practice), and therefore is the preferred option for doing Geographic Data Science.
+- Ensures transparency and reproducibility (which is also the backbone of good scientific practice), and therefore is the preferred option for doing Geographic Data Science.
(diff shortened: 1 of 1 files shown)


09-gis.Rmd
```diff
@@ -240,11 +239,11 @@ get_usage(alg)
...
 ```
 
-Conveniently, the user has not to specify each single parameter.
-In case a parameter is left unspecified, `run_qgis()` will automatically use the corresponding default value as argument if available.
+Conveniently, the user does not need to specify each single parameter.
+In case a parameter is left unspecified, `run_qgis()` will automatically use the corresponding default value as an argument if available.
 To find out about the default values, run `get_args_man()`.  
 
-To finally get rid off the slivers, we specify that all polygons with an area less or equal to 25000 m^2^ should be joined to that neighboring polygon with the largest area (see right panel of Figure \@ref(fig:sliver-fig)).
+To get rid off the slivers, we specify that all polygons with an area less or equal to 25000 m^2^ should be joined to the neighboring polygon with the largest area (see right panel of Figure \@ref(fig:sliver-fig)).
```
Robinlovelace: suggestion: `to remove the slivers` (off is a typo)

Robinlovelace:
Many thanks.

09-gis.Rmd
jannes-m: ok, changed! Thanks!

PR #279, opened 2018-08-20
ideas for application chapters names

Nowosad:
@jannes-m @Robinlovelace What do you think about it?

### DIFF
--- a/12-transport.Rmd
+++ b/12-transport.Rmd
@@ -1,6 +1,6 @@
 # (PART) Applications {-}
 
-# Transport applications {#transport}
+# Transportation {#transport}
 
 ## Prerequisites {-}
 
--- a/14-eco.Rmd
+++ b/14-eco.Rmd
@@ -1,4 +1,4 @@
-# Ecological use case {#eco}
+# Ecology {#eco}
 
 ## Prerequisites {-}
 


jannes-m:
Well, the chapter is on spatial cross validation, for this one needs statistical learning.The glm is just there to gently introduce spatial cv.Per se it is not on modeling. You would not name your transportation chapter something like "Routing" just because routing is an important technique for transportation.Hence,I would name c11 simply spatial cross validation since this is what the chapter is about.Or statistical learning and spatial cross validation

jannes-m:
I wouldn't name the last chapter Ecology, either because it is just one very specific ecological use case,i.e.only a tiny, tiny part of Ecology.

Nowosad:
@jannes-m My thinking is that this book part is called "Application". Therefore, for example chapter 14 is (in my mind) Application: Ecology.

Nowosad:
@jannes-m I suggest just using "Statistical learning" for Ch11.

Robinlovelace:
> @jannes-m I suggest just using "Statistical learning" for Ch11.

I'd be happy with that if Jannes is happy to with this change.

Regarding the ecology chapter, the current chapter name "Ecological use case" is also very broad, no better from that perspective IMO. How about: https://github.com/Robinlovelace/geocompr/pull/279/commits/847585c5956f946f669a9075b1fd1b595e6fd535#diff-5f2534827f08b43d31afe04558257100L1  ?

jannes-m:
"Statistical learning" is fine. Since we are writing a book on geocomputation it should be clear that we apply it to geographic data.

jannes-m:
I agree with you that "Ecological use case" is also rather broad. How about "Predictive mapping of biodiversity". This would also imply that there is a gradient. But maybe Jakub's shortcut idea (Application: Geomarketing/Transportation/Ecology) is also ok, let's talk about it tomorrow, ok?

PR #293, opened 2018-08-27
Propose new chapter names for 3:5

Robinlovelace:


Robinlovelace:
Closing this for now because we've not decided. I think the +s of adding `data` to the titles of c3 and c4 outweigh the costs but don't have a strong preference, can see advantages either way. @jannes-m the reason for adding the word is because `attributes` are a very specific thing in R, e.g. as shown in the reprex below:

``` r
attributes(spData::nz)
#> $names
#> [1] "Name"          "Island"        "Land_area"     "Population"   
#> [5] "Median_income" "Sex_ratio"     "geom"         
#> 
#> $row.names
#>  [1]  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16
#> 
#> $class
#> [1] "sf"         "data.frame"
#> 
#> $sf_column
#> [1] "geom"
#> 
#> $agr
#>          Name        Island     Land_area    Population Median_income 
#>          <NA>          <NA>          <NA>          <NA>          <NA> 
#>     Sex_ratio 
#>          <NA> 
#> Levels: constant aggregate identity
```

Created on 2018-08-29 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

It affects beginners and experienced users. Beginners will likely get confused when they later discover the `attributes()`, which are key to spatial data. Experienced users may think the chapter is about changing the bounding box etc. So @jannes-m and @Nowosad I still suggest we change them. Happy for that? We could try it in the html version and see how it looks.

Robinlovelace:
To make matters more complicated the attributes of sfc objects are also important and can be manipulated: 

``` r
attributes(spData::nz$geom)
#> $class
#> [1] "sfc_MULTIPOLYGON" "sfc"             
#> 
#> $precision
#> [1] 0
#> 
#> $crs
#> $epsg
#> [1] 2193
#> 
#> $proj4string
#> [1] "+proj=tmerc +lat_0=0 +lon_0=173 +k=0.9996 +x_0=1600000 +y_0=10000000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs"
#> 
#> attr(,"class")
#> [1] "crs"
#> 
#> $n_empty
#> [1] 0
#> 
#> $bbox
#>    xmin    ymin    xmax    ymax 
#> 1090144 4748537 2089533 6191874 
#> attr(,"class")
#> [1] "bbox"
```

Created on 2018-08-29 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

jannes-m:
@Robinlovelace Oh, yes, you are right! It is always a struggle with interdisciplinary terminology. In the GIS world, attributes refer to the columns of a dataframe, this is why I suggested the chapter name. But yes, you are right, attribute has a different meaning in the R world. Attribute data operation is probably not a very common term, still it is specific, so I would vote for it now.

Robinlovelace:
Also a strong option. Of the two options:

- 3 Attribute data operations, 4 Spatial data operations 
- 3 Attribute data operations 4 Spatial operations 

The first is more consistent. I suspect the + of consistency outweighs the -s of extra words. Are there any more -s to the latter option? Thoughts @jannes-m also welcome to help decide.

Nowosad:
Ok. Consistency FTW

PR #296, opened 2018-09-01
Rev alg

jannes-m:
Hey Robin,
though I think it is a nice chapter showing how to compute on the language, I have the feeling that it could be written a bit easier. Secondly, I also think the functions you write could be a bit more generic since we should encourage our readers from the beginning to think of the pitfalls other users might face when using these functions. But yes, there might be a tradeoff between easy explanation and a generic way of code writing. Thirdly, I am not sure why everything must be computed from scratch and why you do not want to use sf. I have made several suggestions how to facilitate the reading and how to make the code at times more readable. I hope you find these suggestions valuable. Have a look, and then we can discuss how to further proceed if you like.


10-algorithms.Rmd
```diff
@@ -33,7 +33,8 @@ That is the topic of section \@ref(functions).
 <!-- nor entirely focused on implementations via a GUI or CLI in a particular sofware package (as the first part of this book is, with its focus on implementations in various R packages). -->
 <!-- The focus of this chapter is on understanding, using reproducible code and clear explanation. -->
 The example of finding the centroid of a polygon will be used to tie these concepts together.
-Of course, you already know how to do that with `st_centroid()` having read Chapter \@ref(geometric-operations).
+Of course, you already know how to do that with `st_centroid()` having read Chapter \@ref(geometric-operations). 
+<!-- JM: I think the quote refers to something else, e.g., humans can easily recognize faces or can easily take a step, something which is very hard for a computer/roboter. centroid computation might not be the most complex procedure in the world, still, when asking your students how to do so you will find that they will not find it as easy as recognizing a friend -->
```
Robinlovelace: I think this *is* what the quote is referring to. Here's the quote in extended context:

> Computer software is generally designed to undertake particular tasks or
to solve problems. Geographic Information Systems (GIS) are no excep-
tion, and of course in this case the tasks or problems involve spatial
data. One of the most intriguing things about spatial data problems is
that things which appear to be trivially easy to a human being can be
surprisingly difficult on a computer. This is partly due to the complex
nature of spatial data, and partly to the way that computers tackle
problems.

Happy to consider suggestions to change the prose, but I think what is written is what we mean: devising algorithms is hard.

10-algorithms.Rmd
```diff
@@ -79,6 +80,7 @@ By default RStudio 'code-checks' R scripts and underlines faulty code with a red
 knitr::include_graphics("figures/codecheck.png")
 ```
 
+<!-- JM: the reprex package is great especially for communicating on Github, via twitter, etc. but can you really use it to test if your own code is reproducible?-->
```
Robinlovelace: Yes, but agree it's most useful for communicating to others how to get to a given state and to ask questions or illustrated bugs. That may be worth mentioning. How about changing the first line of the note to read:

```
 A useful tool for reproducibility and communication is the **reprex** package. 
```

Robinlovelace:
Great, thanks for all the comments. **sf** is used, to demonstrate how much easier it is to use existing geoalgorithms than developing your own. There are advantages of using base R to show how algorithms are developed from first principles. Initially I started using the **decido** package for triangulation but, partly based on reviewers comments, I demoted it to an extended example: https://geocompr.github.io/geocompkg/articles/

I don't agree with some of the suggested changes and suggest alterations based on some of your comments - very happy to tweak things to improve the clarity of the chapter.

I just re-started the travis build, not sure why it's failing. Suggestion: get that working and act on the comments-on-comments. Then we can move (or should I say merge) ahead!

Nowosad:
Travis error:
```

Quitting from lines 192-200 (10-algorithms.Rmd) 
Error in library(geometry) : there is no package called 'geometry'
```

Robinlovelace:
I suggest not adding another dependency at this stage.

10-algorithms.Rmd
jannes-m: yeah, ok, then we should mention, that it is relatively easy for a human being to approximately find the centroid of a polygon but that a computer needs a lots of commands to achieve the same result.

10-algorithms.Rmd
jannes-m: Ok

jannes-m:
@Robinlovelace My point is that the reader might get easily distracted from the overarching purpose, computing the centroid which involves four steps, when each of these steps consists of various substeps. Therefore, I suggested to use already available functionality for performing these substeps (even the formulas can remain in the text and you can say, luckily, this was already implemented). Avoiding the usage of another package is not a very strong counter-argument when the text gets easier to digest for the reader through the usage of another package. But if you still think it is better to explain each substep in detail, I would try to vectorize the code and the functions. This would be more R-like (sometimes I got the impression you literally translated Python code), easier to read and more generic.

Robinlovelace:
Happy to see updates that do these things. Could `st_triangulate()` help? Originally I agreed with your approach and used **decido** to do the work of creating triangles. But a reviewer said that that made things complicated. At present my view is that the advantages of outsourcing triangle creation with `ind = delaunayn(poly_mat)` are outweighed by the disadvantages of using another package. It also raises the question: why not use[ **decido**](https://github.com/hypertidy/decido) (which is based on super-fast and modular C++ code and I know works for this) or `st_triangulate()` (which I've not tried for this)?

But happy to be persuaded by the evidence of an updated PR, a build that is working (e.g. by building from a new branch of **geocompkg**) and explanation of the benefits for the reader. @Nowosad what do you reckon?

Robinlovelace:
Good use of `st_triangulate()`. I'm persuaded!

Robinlovelace:
Regarding `lapply()` I think it's good to introduce it here. Regarding a section of algorithms, I think it's good if it uses base R: generally people should not get in the habit of not using compartively bloated tidyverse functions (good for analysis not good for development).  I worry that if we show example algorithms that rely on `dplyr`'s dialect of R (which is a --- wonderfully simple and expressive --- language in its own right for analysing data frames) people will go down that route rather than working things out in the comparatively low level base R language.

jannes-m:
Robin, I think I made some constructive suggestions. Still, I have the feeling you are not entirely convinced (which is ok). Therefore, I suggest that you update the chapter. This is much easier since you will only change it in a way that you like. Afterwards, I am happy to comment and contribute to the updated contents. 

Regarding `lapply()`, of course, I agree with you that one should know how to use it. On the other hand, `group_by()` makes a wonderful job on dataframe objects. Still, we can replace `group_by()` by `aggregate()`, `by()` or alike. One could also use `lapply()` but then we would have to use `split()` and `do.call(rbind, ...)`, i.e. much harder to follow. However, we could add a step 6: using `lapply()` to compute the centroids of more than one polygon.

Robinlovelace:
All sounds good to me and definitely appreciate the constructive suggestions. I am averse to adding more dependencies and wanted this chapter to be a 'base R-y' as possible. That said I can see advantages on the other hand, seeing as readers should be well versed in the tidy verse from c3. On balance I have a mild preference for keeping it as is dependency-wise. I have no problem shifting things around and will consider your comments when I read-through it and respond to incoming reviewer comments, always room for improvement.

The main reason for not acting on all these constructive suggestions at the moment is more mundane however: time (hence closing for now)!

jannes-m:
Ok, let's wait for the reviewers' comments, good idea, would have suggested the same 😊

PR #297, opened 2018-09-05
Rev concl jm

jannes-m:
Hey Robin, 
here, you find some suggestions for the last chapter. One thing, is it necessary that we call it "social benefit"? Can we broaden it a bit? Geocomputation for the common good/for everybody? I mean the applications don't have to have a social touch (though they can). I mean it's really nice that you can use the tools if they are needed, for example in emergency cases/crisis areas, etc.. Makes me think of OSM's https://tasks.hotosm.org/. This is nice but I think it is just one application, therefore, I would broaden the term a bit while one still can point out that it would be nice to use geocomputation for the social benefit. But then, we have to prepare ourselves for questions like, when have you used geocomputation the last time for saving/helping the world/society/poor? Have you?

### DIFF
--- a/15-synthesis.Rmd
+++ b/15-synthesis.Rmd
@@ -92,8 +92,18 @@ We have been selective, emphasizing some topics while omitting others.
 We have tried to emphasize topics that are most commonly needed in real-world applications such as geographic data operations, projections, data read/write and visualization.
 On the other hand, we have omitted topics that are less commonly used, or which are well catered for in other resources.
 Point pattern analysis, spatial interpolation (kriging) and spatial epidemiological modelling, for example, are important topics.
-But there is already excellent material on these things.
+But there is already excellent material on these things such as @baddeley_spatial_2015 and @bivand_applied_2013.
+<!-- one could also add @brunsdon_introduction_2015 -->
 <!-- Todo: add citations to these materials (RL) -->
+Another topic which we barely touched is remote sensing though especially everything related to raster analysis is a good introduction to remote sensing with R.
+If you want to know more, have a look at @wegmann_remote_2016.
+
+Instead of providing the reader with spatial statistical modeling and inference, we mainly chose to present machine-learning algorithms.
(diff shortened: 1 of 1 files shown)


15-synthesis.Rmd
```diff
@@ -55,10 +55,11 @@ The range of options grows further when other R packages are considered: you cou
 We recommend using **sf** and the other packages showcased in this book, for reasons outlined in Chapter \@ref(spatial-class), but it's worth being aware of alternatives and being able to justify your choice of software.
 
 We deliberately covered both **tidyverse** and base R approaches to attribute data operations.
-Chapter \@ref(attr) showed how both `nz["Name"]` and `nz %>% select(Name)` can be used to achieve the same result, for example.
+Chapter \@ref(attr) showed how both `nz[, "Name"]` and `nz %>% select(Name)` can be used to achieve the same result, for example.
```
Nowosad: @jannes-m The previous version was probably better. 
- On sf object, both `nz["Name]` and `nz[, "Name"]` give the same result
- However on data.frame objects `nz_df["Name"]` is the same as `nz_df %>% select(Name)`
- `nz_df[, "Name"]` is more like `nz_df %>% pull(Name)`

jannes-m: Robin, while on the bike today,what do you think about replacing "Geocomputation for social benefit" by "Advocating open source software"? This is broader, we all agree, and you can still say that using an open source approach, social benefits are more likely to happen.

Nowosad: @jannes-m `nz["Name"]` and `nz[, "Name"]` give the same output - an sf object with two columns. This is due to "stickiness" of a geometry column.

Robinlovelace:
Will aim to look at this this evening.

jannes-m:
@Nowosad yeah, I got this. I was talking of the `data.frame` case.

jannes-m:
:+1:

PR #304, opened 2018-09-20
Mds ch10 mods

mdsumner:
Two minor commits

1) to reword an awkward sentence (too casual)
2) suggest packages on CRAN for polygon triangulation (mine ... they wrap external libs)

Great chapter!

### DIFF
--- a/10-algorithms.Rmd
+++ b/10-algorithms.Rmd
@@ -33,8 +33,7 @@ That is the topic of section \@ref(functions).
 <!-- nor entirely focused on implementations via a GUI or CLI in a particular sofware package (as the first part of this book is, with its focus on implementations in various R packages). -->
 <!-- The focus of this chapter is on understanding, using reproducible code and clear explanation. -->
 The example of finding the centroid of a polygon will be used to tie these concepts together.
-Of course, you already know how to do that with `st_centroid()` having read Chapter \@ref(geometric-operations).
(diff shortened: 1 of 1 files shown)


10-algorithms.Rmd
```diff
@@ -148,7 +147,7 @@ It helps to further break-down this approach into discrete tasks before writing
 <!-- 5. Return the result -->
 
 These steps may sound straightforward, but converting words into working code requires some work and plenty of trial-and-error, even when the inputs are constrained.^[
-The algorithm will only work for *convex polygons*, which contain no internal angles greater than 180° --- no star shapes allowed.
+The algorithm will only work for *convex polygons*, which contain no internal angles greater than 180° --- no star shapes allowed. See packages **decido** or **sfdct** for approaches to non-convex polygons using external libraries.
```
Robinlovelace: Thanks for fixing that big omission. We should also add here a link to this (in progress) vigette that shows how you can make those triangles way faster with decido: https://geocompr.github.io/geocompkg/articles/algorithm.html

The result of rewriting the chapter: initially it used decido but a reviewer rightly pointed out that added complexity and increased the probability of confusion! This comment is the ideal place. Only question: does this deserve to be in the main text? I suspect so.

mdsumner:
Oh nice!  I was impressed by with the example it's a good unpacking of something that is v hard in the general case but is composed of important techniques

PR #305, opened 2018-09-20
Include frontcover

marcosci:
I included the code in `code` and the cover in `images`.
I also put the console idea of Robin in one of the tiles - opinions?


Nowosad:
Thanks @marcosci - I am going to merge it now to improve the code style a bit. 
One question - what do you think about removing this vertical line in the Robin's console idea?

Robinlovelace:
Also looking at this.

Note I had errors in 2 places:

```r
r <- raster("images/r_logo.tif")
# and:
  mutate(id = ifelse(r_logo==255, 0, 1)) %>% 
```

marcosci:
@Nowosad sorry for that :yum: and sure, if Robin is fine with that remove it :+1: 

@Robinlovelace what are the errors?

Robinlovelace:
May be because I ran it from different directory. Happy for the line to go. Also suspect it needs to be higher resolution but looking great to me.

Nowosad:
I've cleaned the code a few minutes ago. 
Everything ran well on my laptop.

PR #307, opened 2018-09-23
dont run final code chunk

Robinlovelace:
Heads-up: this fixes the build, on my computer at least.

The final chunk generates an error - fixing the code chunk and still running it is preferable.


jannes-m:
I don't get it, I just ran the whole again, everything works fine. The beauty of this code is that it returns an interactive leaflet map which is nice in the case of the online book. However, if you convert all interactive maps into static maps, then we can also set `eval=FALSE`  here, and include the corresponding graphic in a separate code chunk.

Robinlovelace:
Yeah the code is awesome, we just need to find a way to make it run. This PR was a test to see if it will make travis happy. Seems so.

PR #309, opened 2018-09-26
Added draft foreword

rsbivand:
Please let me know what you think - apart from asking me to shorten sentences (I'm not good at that!).


Robinlovelace:
I think this looks great! Many thanks, will be in touch re. detail.

PR #312, opened 2018-09-27
Revised draft foreword

rsbivand:
Minor changes taking into account possible changes in conclusion (maybe drop reference to conclusion in parentheses if it becomes irrelevant),


Robinlovelace:
Many thanks Roger!

PR #313, opened 2018-09-29
Reframe geocomputation in c1 and c15

Robinlovelace:
Put in as a PR to provide time to discuss these changes if needs be. Thoughts from anyone welcome, especially @rsbivand.

### DIFF
--- a/01-introduction.Rmd
+++ b/01-introduction.Rmd
@@ -37,18 +37,16 @@ This may sound simple and easy to achieve (which it is if you carefully maintain
 
 ## What is geocomputation?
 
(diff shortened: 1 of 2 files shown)


rsbivand:
Looks OK to me, but conditional on the difficulty of reading the green sections alone, without the rest of the text in those chapters.

Robinlovelace:
Great, thanks for the feedback. There's still plenty of scope for further change and hopefully improvement pending the final technical review comments. I think these changes address some of those comments and make part of the foreword out of date.

Robinlovelace:
Merging following :+1: from @Nowosad + @jannes-m thanks for the feedback!

PR #315, opened 2018-10-01
 revised foreword to suit ch 1 & 15 changes 

rsbivand:
Hope this tunes the draft foreword to match changes in chs 1 & 15


jannes-m:
@rsbivand I can only echo Robin's words!! Thank you so much for this beautiful foreword!

PR #322, opened 2018-10-09
Add web services section

Robinlovelace:


07-read-write-plot.Rmd
```diff
@@ -136,6 +136,35 @@ world2 = spData::world
 world3 = st_read(system.file("shapes/world.gpkg", package = "spData"))
 ```
 
+## Geographic web services
+
+In an effort to standardise web APIs for accessing spatial data, the Open Geospatial Consortium (OGC) has created a number of specifications for web services (collectively known as [OWS](http://www.opengeospatial.org/standards/owc)).
```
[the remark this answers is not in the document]
Robinlovelace: Yes and yes, put in as a PR to get feedback. More broadly, is this new section useful/wanted? I think so but think it would need to be longer so more content welcome. In other words is this a useful code chunk and where next after demonstrating `GetCapabilities`?:

``` r
u = "http://geodata.nationaalgeoregister.nl/bag/wfs?"
g = "GetCapabilities"
res = httr::GET(url = u, query = list(request = g))
txt = httr::content(res, "text")
#> No encoding supplied: defaulting to UTF-8.
xml2::read_xml(txt)
#> {xml_document}
#> <WFS_Capabilities version="2.0.0" schemaLocation="http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd http://inspire.ec.europa.eu/schemas/inspire_dls/1.0 http://inspire.ec.europa.eu/schemas/inspire_dls/1.0/inspire_dls.xsd" updateSequence="1913" xmlns:wfs="http://www.opengis.net/wfs/2.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.opengis.net/wfs/2.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:gml="http://www.opengis.net/gml/3.2" xmlns:fes="http://www.opengis.net/fes/2.0" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:inspire_dls="http://inspire.ec.europa.eu/schemas/inspire_dls/1.0" xmlns:inspire_common="http://inspire.ec.europa.eu/schemas/common/1.0" xmlns:bag="http://bag.geonovum.nl">
#> [1] <ows:ServiceIdentification>\n  <ows:Title>BAG</ows:Title>\n  <ows:Ab ...
#> [2] <ows:ServiceProvider>\n  <ows:ProviderName>PDOK</ows:ProviderName>\n ...
#> [3] <ows:OperationsMetadata>\n  <ows:Operation name="GetCapabilities">\n ...
#> [4] <FeatureTypeList>\n  <FeatureType>\n    <Name>bag:ligplaats</Name>\n ...
#> [5] <fes:Filter_Capabilities>\n  <fes:Conformance>\n    <fes:Constraint  ...
```

<sup>Created on 2018-10-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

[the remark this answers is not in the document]
Nowosad: @Robinlovelace truth to be told - I am not a regular user of these webservices, so I have no idea what should we show here.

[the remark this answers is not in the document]
Robinlovelace: Know anyone else who is?

jannes-m:
I don't want to interfere too much since I am far away. But so far the example just shows a simple GET request which is not really specific to geographic data. Additionally, it does not download any geographic data. So maybe it would be helpful to add a code chunk which actually downloads some geographic data into R (would then also better fit the chapter title)? Please also remember that leaflet, mapview and tmap provide access to various web mapping services, maybe worth mentioning here (+ cross reference to the relevant sections)?

Robinlovelace:
Thanks for the input. Yes indeed an example that actually downloads some data is needed.

Robinlovelace:
I think we could end-up going off on a tangent here. There are packages for working with OWS in R but all are under development. I've mentioned three briefly, mentioned leaflet and linked to appropriate further reading resources. I think that is sufficient for now. Plan to merge this tomorrow so please let me know if anyone has further comments, especially @cvitolo. Thanks again for raising this issue. Finally we have a section that I'd be happy to see in print!

07-read-write-plot.Rmd
```diff
@@ -136,6 +136,62 @@ world2 = spData::world
...
+#> [3] <ows:OperationsMetadata>\n  <ows:Operation name= ...
+#> [4] <FeatureTypeList>\n  <FeatureType>\n    <Name>ba ...
+#> [5] <fes:Filter_Capabilities>\n  <fes:Conformance>\n ...
+```
+
+The previous example shows how to gain low-level access to web services using **httr**, which can be useful for understanding how web services work.
+For many everyday tasks, however, a higher-level interface may be more appropriate, and a number of R packages, and tutorials, have been developed precisely for this purpose, as outlined below (an in-depth coverage of OWS services is beyond the scope of the book).
+
+Packages **ows4R**, **rwfs** and **sos4R** have been developed for working with OWS services in general, WFS and the sensor observation service (SOS) respectively.
+As of October 2018 all of them are still works in-progress and only one (**ows4R**) is on CRAN.
```
Robinlovelace: Heads-up @eblondel we've mentioned your pkg in our open source book! Please take a look at mention of it and let me know if you have comments about our description of **ows4r** (looks like an interesting project) or the section in general.

jannes-m:
@Robinlovelace it took me quite some time to figure out how to work with **ows4r**. One reason for this was that I don't use WFS on a regular basis, another reason was certainly that the package documentation could be a bit more user-friendly. There are a few other shortcomings, please have a look at my commits, I have listed them there. Still, I think it would be be great if we included the US national parc or US airport WFS download, and carefully mention that the package and related ones are still pretty much under development.

07-read-write-plot.Rmd
eblondel: Thanks @Robinlovelace for mentioning ows4R in your book, indeed it's great!

I want to be very clear first on some aspect: ``ows4R`` is a new package, and giving its scope (OGC web-services), you can assume this will be "in development" for a lot of time :-), because it is expected to grow, and grow more.  This doesn't prevent that ows4R is going into release lifecycle with CRAN, and this means that stable software is expected to be released (of course we can still find bugs, it's quite natural). If we stop dev on it, this would be rather a bad news, and backward compability will be ensured unless a major enhancement would need user code adapation.

For your book, I would suggest to identifies what does OWS standards for, earlier in the text (and if not yet done.), ie OGC Web-Services, (with a link the OGC standards web-page, maybe?)

Maybe you could also mention the support of the Catalogue Service for the Web (CSW) standard, which targets metadata management, with binding with the [geometa](https://github.com/eblondel/geometa) package, for writing and reading of ISO/OGC geographic metadata.  FYI, these 2 packages are under active development. For what concerns ``ows4R``, any OGC standard service is potentially candidate for integration into ows4R, with priority given to standards that are not supported at all yet in R (while some other standard web-service have some implementations available like sos4R). A prototype for WCS (Web Coverage Service) is under investigation with mapping to the raster package, not that easy, WCS is much more complex that the simple WFS, this will target data accessing in multi-dimensional geographic data. I hope that a first client could be released next year.

If you want more information, i'm available.

eblondel:
For what concerns documentation, @jannes-m don't know if you sticked with the Rdoc, but I would suggest that you get into the ows4R wiki page that provides documentation how to use the services: https://github.com/eblondel/ows4R/wiki, I'm the author so my view is not objective, and as for all emerging software, the documentation will also evolve, but so far the feedback I received from the users is that they were capable to use both WFS and CSW without my help, just looking at the wiki and the examples provided. But if you have suggestions for making the doc more user-friendly, they are welcome.

07-read-write-plot.Rmd
Robinlovelace: Many thanks for the detailed comment!

Robinlovelace: To response to some of your suggestions, we do state what OWS early so that's good and link to the page and other resources later on. We could go into more detail. But the most important thing in this section I think, is that it exists! What I mean by that is that in a chapter on data I/O it's worth flagging all these (awesome) standard services. Because the R-OWS interfaces do not seem to be advanced the main priority is to point people in the right direction in a future proof way.

It sounds like `owsR` is stable and future-proof. Would it be possible to use it to download some features from a web service, building on the following reproducible example?

``` r
base_url = "http://geodata.nationaalgeoregister.nl/bag/wfs"
q = list(request = "GetCapabilities")
res = httr::GET(url = base_url, query = q)
res$url
#> [1] "http://geodata.nationaalgeoregister.nl/bag/wfs?request=GetCapabilities"
txt = httr::content(res, "text")
#> No encoding supplied: defaulting to UTF-8.
xml2::read_xml(txt)
#> {xml_document}
#> <WFS_Capabilities version="2.0.0" schemaLocation="http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd http://inspire.ec.europa.eu/schemas/inspire_dls/1.0 http://inspire.ec.europa.eu/schemas/inspire_dls/1.0/inspire_dls.xsd" updateSequence="1913" xmlns:wfs="http://www.opengis.net/wfs/2.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.opengis.net/wfs/2.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:gml="http://www.opengis.net/gml/3.2" xmlns:fes="http://www.opengis.net/fes/2.0" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:inspire_dls="http://inspire.ec.europa.eu/schemas/inspire_dls/1.0" xmlns:inspire_common="http://inspire.ec.europa.eu/schemas/common/1.0" xmlns:bag="http://bag.geonovum.nl">
#> [1] <ows:ServiceIdentification>\n  <ows:Title>BAG</ows:Title>\n  <ows:Ab ...
#> [2] <ows:ServiceProvider>\n  <ows:ProviderName>PDOK</ows:ProviderName>\n ...
#> [3] <ows:OperationsMetadata>\n  <ows:Operation name="GetCapabilities">\n ...
#> [4] <FeatureTypeList>\n  <FeatureType>\n    <Name>bag:ligplaats</Name>\n ...
#> [5] <fes:Filter_Capabilities>\n  <fes:Conformance>\n    <fes:Constraint  ...
```

<sup>Created on 2018-10-18 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Or at least point us in the right direction (will look at your other comment also, very informative). Note: we don't want to make this section too long!

jannes-m:
@eblondel Thanks for all your comments and input! Thank you also for working on the OGC web services for R!
Regarding the documentation, in fact I have read the wiki page. IMO a reproducible example could be helpful, especially for an audience who has no or only little experience with web services. 
Additionally, I had a few problems with German/Dutch WFS (will show in an reproducible example). I also couldn't figure out if it was possible to provide an additional namespace parameter to `getFeatures()` if necessary.

eblondel:
``namespace`` is AFAIK not a vendor parameter of WFS getFeatures operation, so you will not find it. The getFeatures method let you specifying whatever vendor parameter supported by the WFS spec, like ``cql_filter``. If you refer to namespace as used in GeoServer WFS, this is product specific, either you have to provide it initially within the base URL ( http://myserver/geoserver/mynamespace/wfs), or as prefix of the layer name when finding feature type or getting features

jannes-m:
@eblondel Thanks for the heads-up, will try this with this namespaces!

eblondel:
I take the exemple of the UN-FAO geoserver, that i've been managing. In geoserver either you do not specify any namespace in the URL, in that case you call a "global" instance of the WFS (all namespace):

```{r}
wfs1 <- WFSClient$new("http://www.fao.org/figis/geoserver/wfs", serviceVersion = "1.0.0", logger = "INFO")
sf1 <- wfs1$getFeatures("fifao:UN_CONTINENT2")
```

You see that I specified a prefix "fifao" (which is the default namespace) on some layer. Geoserver identifies its layers handling the namespace as prefix, so it's part of the actual identifier. 

And here you will note i've also call getFeatures directly from the WFS client, which is a short cut, so you don't have to find the layer first from the capabilities, and then get Features, you can do it once.

In other case, you can specify the namespace in the base URL. :

```{r}
wfs2 <- WFSClient$new("http://www.fao.org/figis/geoserver/fifao/wfs", serviceVersion = "1.0.0", logger = "INFO")
sf2 <- wfs2$getFeatures("fifao:UN_CONTINENT2")
```
This has an advantage when you try to access a big geoserver with many namespaces, and a huge number of layers, it's better to call a sub-instance of the WFS service, targeting the namespace you want to explore.

As for extra parameters, an example of the cql filter:

```{r}
wfs <- WFSClient$new("http://www.fao.org/figis/geoserver/area/wfs", serviceVersion = "1.0.0", logger = "INFO")
sf <- wfs$getFeatures("area:FAO_AREAS", cql_filter = URLencode("F_CODE= '27'"))
```
Hope this helps

Robinlovelace:
Many thanks for the examples @eblondel. Following it I've changed the example to use your FAO example which has a shorter url and global coverage. To summarise, the code in the chunks is now:

```r
base_url = "http://www.fao.org/figis/geoserver/wfs"
q = list(request = "GetCapabilities")
res = httr::GET(url = base_url, query = q)
res$url
txt = httr::content(res, "text")
xml2::read_xml(txt)

library(ows4R)
wfs = WFSClient$new("http://www.fao.org/figis/geoserver/wfs",
                      serviceVersion = "1.0.0", logger = "INFO")
areas_all = wfs$getFeatures("area:FAO_AREAS")
area_27 = wfs$getFeatures("area:FAO_AREAS",
                      cql_filter = URLencode("F_CODE= '27'"))
```

Couple of questions: what URL would be used to get the same result using `httr::GET()`? And how can one find the namespaces such as `area`?

Very grateful for this input, as you can see we're new to this. Awesome that it can return `sf` objects directly.

eblondel:
Hi @Robinlovelace, the namespaces are specific workspaces set-up at Geoserver level, they are not part of the WFS standard, so in principle following the OGC specs, there is no way to list the workspaces in Geoserver. This is really a specific feature of Geoserver, consider that other geographic server products have similar but different mechanism.  For example Thredds handles nested directories for namespaces, comparable to a file system. The only way you have to "discover" the different namespaces in geoserver is to query services based on the global namespace, in the above example ``http://www.fao.org/figis/geoserver/wfs``.

In case of WFS the above code involves three kinds of standard requests:
* GetCapabilities: http://www.fao.org/figis/geoserver/wfs?request=GetCapabilities&version=1.0.0 (to getservice apabilities)
* DescribeFeatureType: http://www.fao.org/figis/geoserver/wfs?request=DescribeFeatureType&version=1.0.0&typeName=area:FAO_AREAS (to get schema description)
* GetFeatures: http://www.fao.org/figis/geoserver/wfs?request=GetFeature&version=1.0.0&typeName=area:FAO_AREAS (to get data)

Note that the way requests are built varies according to the service version. The above links are then specific to WFS version 1.0.0. ows4R manages WFS versions, so the user doesn't have to know the WFS specification, specifying the version is enough, ows4R will take of building requests.

I'm not sure that using this particular geoserver for the book is a good idea, since I cannot exclude this particular Geoserver URL will be maintained in time (at short-term surely, but not at medium term), i would rather recommend to use a demo online Geoserver (if enough stable in time), ideally I would consider to deploy a simple demo geoserver for the purpose of the book (the best in my opinion as you would control its availability). Geoserver installation comes with some sample data you can use to perform illustrative requests.

PS: WFS is not only download data but also allows performing transactions (Insert, Update, Delete operations)  The current version of ows4R doesn't support yet the WFS Transaction mode (but it does for CSW!). I hope I could add it early next year. If you are interested into spatial data upload from R to Geoserver, you may want to look at [geosapi](http://github.com/eblondel/geosapi/wiki)

Hope this helps

Robinlovelace:
Hi @eblondel I've fixed the typo in FAO - your use of a fisheries example confused me.

Robinlovelace:
I've updated the section so it now shows importing data with the intermediate step of `GET()` and directly with the **ows4R** package and links to useful resources. That's enough I think so will merge. Further comments/suggestions welcome. Glad to have this section added at the last minute, will be useful for many people I think.

jannes-m:
@eblondel I have one more question since you are a web service expert and pretty familiar with the UN-FAO geoserver -:). I can asses the typenames using **XML** and **RCurl** as follows:

```r
library("XML")
library("RCurl")

doc = xmlParse("http://www.fao.org/figis/geoserver/wfs?request=GetCapabilities")
root = xmlRoot(doc)
# getting familiar with the structure of the XML document
names(root)
names(root[["FeatureTypeList"]])
root[["FeatureTypeList"]][["FeatureType"]][["Name"]]
# and now retrieve all names
xmlSApply(root[["FeatureTypeList"]], function(x) xmlValue(x[["Name"]]))
```

And normally it should be even easier using `xpathSApply()`, however, this only returns an empty list:

```r
xpathSApply(root[["FeatureTypeList"]], "//FeatureType//Name", xmlValue)
```

What am I missing?

eblondel:
Note sure to understand why you want to go directly with XML and RCurl? You can do that with ows4R, simply by doing

sapply(wfs$getCapabilities()$getFeatureTypes(), function(x){x$getName()})
wfs, being the WFSClient you created initially.

OGC Web-services return XML responses that are made of different XML namespaces. What you are trying to do is to parse XML but without specifying namespaces of the nodes you are looking for, this is generally the reason why we get empty lists dealing with XML Xpath parsing.

The problem with your example is that your adhoc XML parsing will be specific to OGC service versions (capabilities response structures are different from one version to another), in the same way the way of constructing URL requests. Frankly speaking, I'm not sure to see the benefit of adding adhoc XML parsing examples, unless for people that want to study in depth OGC services. ows4R has been kicked-off precisely to provide this abstraction, and to avoid that people have to get into the OGC XML stuff, and just stick with the standard operations described in the OGC specs.

An additional note on ows4R, for the time being, I didn't add pretty printing of the objects and their nested properties (like FeatureType, name, bounding box, etc), so still the rough R6 object default printing which doesn't facilitate quick printing of ows4R objects, but FYI this is will be done with the next version of the package, in the same way as the pretty print done for geometa objects (you can try it to see how it looks). 

(sorry if you receive this message twice, don't know why but Github is not working properly here today....)

eblondel:
Note sure to understand why you want to go directly with XML and RCurl? You can do that with ows4R, simply by doing

sapply(wfs$getCapabilities()$getFeatureTypes(), function(x){x$getName()})
wfs, being the WFSClient you created initially.

OGC Web-services return XML responses that are made of different XML namespaces. What you are trying to do is to parse XML but without specifying namespaces of the nodes you are looking for, this is generally the reason why we get empty lists dealing with XML Xpath parsing.

The problem with your example is that your adhoc XML parsing will be specific to OGC service versions (capabilities response structures are different from one version to another), in the same way the way of constructing URL requests. Frankly speaking, I'm not sure to see the benefit of adding adhoc XML parsing examples, unless for people that want to study in depth OGC services. ows4R has been kicked-off precisely to provide this abstraction, and to avoid that people have to get into the OGC XML stuff, and just stick with the standard operations described in the OGC specs.

An additional note on ows4R, for the time being, I didn't add pretty printing of the objects and their nested properties (like FeatureType, name, bounding box, etc), so still the rough R6 object default printing which doesn't facilitate quick printing of ows4R objects, but FYI this is will be done with the next version of the package, in the same way as the pretty print done for geometa objects (you can try it to see how it looks).

jannes-m:
Thanks @eblondel! This was just out of curiosity how to do this "manually". I did not plan to include this in the book.

eblondel:
Note sure to understand why you want to go directly with XML and RCurl? 
You can do that with ows4R, simply by doing

sapply(wfs$getCapabilities()$getFeatureTypes(), function(x){x$getName()})

wfs, being the WFSClient you created initially.

OGC Web-services return XML responses that are made of different XML 
namespaces. What you are trying to do is to parse XML but without 
specifying namespaces of the nodes you are looking for, this is 
generally the reason why we get empty lists dealing with XML Xpath parsing.

The problem with your example is that your adhoc XML parsing will be 
specific to OGC service versions (capabilities response structures are 
different from one version to another), in the same way the way of 
constructing URL requests. Frankly speaking, I'm not sure to see the 
benefit of adding adhoc XML parsing examples, unless for people that 
want to study in depth OGC services. ows4R has been kicked-off precisely 
to provide this abstraction, and to avoid that people have to get into 
the OGC XML stuff, and just stick with the standard operations described 
in the OGC specs.

An additional note on ows4R, for the time being, I didn't add pretty 
printing of the objects and their nested properties (like FeatureType, 
name, bounding box, etc), so still the rough R6 object default printing 
which doesn't facilitate quick printing//of /ows4R/ objects, but FYI 
this is will be done with the next version of the package, in the same 
way as the pretty print done for /geometa/ objects (you can try it to 
see how it looks).


Le 22/10/2018 à 12:36, Jannes Muenchow a écrit :

PR #336, opened 2018-12-13
Copyedit c2

Robinlovelace:
Heads-up: I think this should be merged quick because it contains fixes in other chapters. low level becomes correctly low-level throughout. Please review @Nowosad and @jannes-m.


Robinlovelace:
Thanks @jannes-m for the suggestions. I've acted on both of those. Please merge. Great internal review process.

PR #337, opened 2018-12-14
Copyedit c5

jannes-m:
Hey there,
I am not entirely done with the copy-editing of c5 (will let you know when I am done) but I am already creating this PR to avoid that I forget to ask a question.

### DIFF
--- a/05-geometry-operations.Rmd
+++ b/05-geometry-operations.Rmd
@@ -219,7 +219,7 @@ nz_sfc = st_geometry(nz)
 
 Shifting moves every point by the same distance in map units.
 It could be done by adding a numerical vector to a vector object.
-For example, the code below shifts all y-coordinates by 100,000 meters to the north but leaves the x-coordinates untouched (left panel on the Fig. \@ref(fig:affine-trans)). 
+For example, the code below shifts all y-coordinates by 100,000 meters to the north but leaves the x-coordinates untouched (left panel on the Figure \@ref(fig:affine-trans)). 
 
 ```{r}
 nz_shift = nz_sfc + c(0, 100000)
@@ -235,7 +235,7 @@ nz_scale0 = nz_sfc * 0.5
 ```
 
 Local scaling treats geometries independently and requires points around which geometries are going to be scaled, e.g. centroids.
-In the example below, each geometry is shrunk by a factor of two around the centroids (central panel on the Fig. \@ref(fig:affine-trans)).
+In the example below, each geometry is shrunk by a factor of two around the centroids (central panel on the Figure \@ref(fig:affine-trans)).
 To achieve that, each object is firstly shifted in a way that its center has coordinates of `0, 0` (`(nz_sfc - nz_centroid_sfc)`). 
 Next, the sizes of the geometries are reduced by half (`* 0.5`).
 Finally, each object's centroid is moved back to the input data coordinates (`+ nz_centroid_sfc`). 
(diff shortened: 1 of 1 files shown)


05-geometry-operations.Rmd
```diff
@@ -22,23 +22,23 @@ After reading it --- and attempting the exercises at the end --- you should unde
 Section \@ref(geo-vec) covers transforming vector geometries with 'unary' and 'binary' operations.
 <!-- TODO: add something on n-ary ops (RL) -->
 Unary operations work on a single geometry in isolation.
-This includes simplification (of lines and polygons), the creation of buffers and centroids, and shifting/scaling/rotating single geometries using 'affine transformations' (sections \@ref(simplification) to \@ref(affine-transformations)).
+This includes simplification (of lines and polygons), the creation of buffers and centroids, and shifting/scaling/rotating single geometries using 'affine transformations' (Sections \@ref(simplification) to \@ref(affine-transformations)).
```
Robinlovelace: I already changed sections to Sections in all chapters in the book. General rules can be implemented on the who book with a tool like regexxer. But definitely no harm in this.

05-geometry-operations.Rmd
```diff
@@ -502,12 +501,17 @@ names(sfs_st_cast) = abbreviate_geomtypes(names(sfs_st_cast))
 names(sfs_st_cast)[1] = ""
 knitr::kable(sfs_st_cast,
```
jannes-m: @Nowosad pls check if you are happy with the updated table (I think you have created it originally). I am not sure if I understand the columns. Have you chosen randomly the numbers for the `POI` column or where do these numbers come from? Maybe I am missing something, if not, we should clarify the origin of these numbers.

05-geometry-operations.Rmd
```diff
@@ -477,15 +477,14 @@ t2 = cast_all(polyg)
...
+As a result of this process, multi-objects are split into many non-multi-objects.
 
 Table \@ref(tab:sfs-st-cast) shows possible geometry type transformations on simple feature objects.
-(Note the abbreviated geometry type names: POI, MPOI, LIN and GC for example refer to POINT, MULTIPOINT, LINESTRING and GEOMETRYCOLLECTION respectively.)
 Each input simple feature object with only one element (first column) is transformed directly into another geometry type.
 Several of the transformations are not possible, for example, you cannot convert a single point into a multilinestring or a polygon (so the cells `[1, 4:5]` in the table are NA).
-On the other hand, some of the transformations are splitting the single element input object into multi-element one.
-You can see that, for example, when you cast a multipoint consisting of five pairs of coordinated into a point.
+On the other hand, some of the transformations are splitting the single element input object into a multi-element object.
+You can see that, for example, when you cast a multipoint consisting of five pairs of coordinates into a point.
```
jannes-m: @Nowosad To which row in the table are your referring to? 2nd row? But here the POI version only consists of four points, or am I just not understanding the table?

05-geometry-operations.Rmd
jannes-m: ah, good to know! Still, worth to have a look in each chapter because sometimes we have forgotten to write Section and just written `\@ref(...)`

jannes-m:
So, I guess I should move the footnote into the caption of the table. And then we should tell Michele that it is not really possible to use table footnotes?

Robinlovelace:
@jannes-m your suggestion regarding footnotes sounds good. Keep it simple is my suggestion.

PR #339, opened 2018-12-14
improves ch3

Nowosad:


03-attribute-operations.Rmd
```diff
@@ -27,13 +27,12 @@ The name is an *attribute* of the feature (to use Simple Features terminology) t
 Another example is the elevation value (attribute) for a specific grid cell in raster data.
 Unlike the vector data model, the raster data model stores the coordinate of the grid cell indirectly, meaning the distinction between attribute and spatial information is less clear.
 To illustrate the point, think of a pixel in the 3^rd^ row and the 4^th^ column of a raster matrix.
-It's spatial location is defined by its index in the matrix: move from the origin four cells in the x direction (typically east and right on maps) and three cells in the y direction (typically south and down).
+Its spatial location is defined by its index in the matrix: move from the origin four cells in the x direction (typically east and right on maps) and three cells in the y direction (typically south and down).
```
Robinlovelace: Definite improvement.

03-attribute-operations.Rmd
```diff
@@ -638,10 +604,10 @@ data("grain")
 
 ```{block2 type='rmdnote'}
 `raster` objects can contain values of class `numeric`, `integer`, `logical` or `factor`, but not `character`.
-To use character values they must first be converted into an appropriate class, for example using the function `factor()`. 
+To use character values, they must first be converted into an appropriate class, for example using the function `factor()`. 
 The `levels` argument was used in the preceding code chunk to create an ordered factor:
 clay < silt < sand in terms of grain size.
-See the [Data structures](http://adv-r.had.co.nz/Data-structures.html) chapter of @wickham_advanced_2014 for further details on classes.
+See the Data structures chapter of @wickham_advanced_2014 for further details on classes.
```
Robinlovelace: Good suggestion - reference is enough.

PR #342, opened 2018-12-14
improves ch6

Nowosad:


06-reproj.Rmd
```diff
@@ -14,21 +14,16 @@ library(spDataLarge)
...
 ## Introduction
 
-<!-- A vital type of geometry transformation is *reprojecting* from one coordinate reference system (CRS) to another. -->
-<!-- Because of the importance of reprojection, introduced in Chapter \@ref(spatial-class), and the fact that it applies to raster and vector geometries alike, it is the topic of the first section in this chapter. -->
-
 Section \@ref(crs-intro) introduced coordinate reference systems (CRSs) and demonstrated their importance.
 This chapter goes further.
 It highlights issues that can arise when using inappropriate CRSs and how to *transform* data from one CRS to another.
 
-As illustrated in Figure \@ref(fig:vectorplots), there are two types of CRS: *geographic* ('lon/lat', with units in degrees longitude and latitude) and *projected* (typically with units of meters from a datum).
```
Robinlovelace: I don't think this is an improvement. "There are two types of apple" is surely better than "there are two types of apples". Unless I'm missing something, the same should apply to types of Coordinate Reference System (CRS).

06-reproj.Rmd
```diff
@@ -174,23 +168,23 @@ st_distance(london2, london_proj)
...
 WGS84 is the most common CRS in the world, so it is worth knowing its EPSG code: 4326.
 This 'magic number' can be used to convert objects with unusual projected CRSs into something that is widely understood.
 
 What about when a projected CRS is required?
-In some cases it is not something that we are free to decide:
+In some cases, it is not something that we are free to decide:
 "often the choice of projection is made by a public mapping agency" [@bivand_applied_2013].
 This means that when working with local data sources, it is likely preferable to work with the CRS in which the data was provided, to ensure compatibility, even if the official CRS is not the most accurate.
-The example of London was easy to answer because a) the CRS 'BNG' (with its associated EPSG code 27700) is well-known and b) the original dataset (`london`) already had that CRS.
+The example of London was easy to answer because (a) the CRS 'BNG' (with its associated EPSG code 27700) is well known and (b) the original dataset (`london`) already had that CRS.
```
Robinlovelace: I like this new style of lists suggested by the reviewer.

06-reproj.Rmd
Robinlovelace: @Nowosad looking into it - I take that back. I think it is an improvement so keeping it.

PR #344, opened 2018-12-14
Fix cross reference

giocomai:
Cross-reference does not work, I suppose this is the section it is supposed to refer to


Nowosad:
Thanks a lot @giocomai!

PR #355, opened 2018-12-21
copyedit_c13

jannes-m:


jannes-m:
Page 285 of the copyedited pdf: In the header it says "13.4 Tidy the input data". It should state either "13.3 Tidy the input data" or "13.4 Create census rasters". Both headers can be found on page 285, the mixture "13.4 Tidy the input data" is awkward. How can we tell Latex not to do so?

Nowosad:
The similar problem occurs several times in the book, e.g. page 129, 133, 191... 
I have no idea why that happens, but we should work on it after the all other improvements are implemented...

13-location.Rmd
```diff
@@ -498,7 +498,7 @@ reclass = dropLayer(reclass, "pop")
 ```
 
 In common with other data science projects, data retrieval and 'tidying' have consumed much of the overall workload so far.
-With clean data the final step, calculating a final score by summing up all raster layers, can be accomplished in a single line.
+With clean data the final step --- calculating a final score by summing up all raster layers --- can be accomplished in a single line.
```
jannes-m: @Robinlovelace pls check if the sentence is clear and understandable, the copyeditor suggested to change it to:

`With clean data as the final step, calculating a final score by summing up all raster layers can be accomplished in a single line.`
Which is clearly not, what I meant.

Robinlovelace: Looks good except I think there should be an additional comma after 'With the data clean', and removal of the the word 'up' which I think adds little/nothing to the meaning, e.g.:

```
With clean data, the final step --- calculating a final score by summing all raster layers --- can be accomplished in a single line of code.
```

And `line of code` is clearer than `line`.


 Good use of ---!

PR #362, opened 2019-01-14
Update 11-spatial-cv.Rmd explanation

jbixon13:
Added small change for clarity -- clustering is generally an unsupervised learning technique. 

https://towardsdatascience.com/clustering-unsupervised-learning-788b215b074b


jbixon13:
"As cluster analysis is unsupervised," - quoted from mlr documentation @ https://arxiv.org/pdf/1609.06146.pdf

Clarified that mlr is a package for both supervised and unsupervised learning.

jannes-m:
Thanks, the clarification is very much welcome, especially in this sentence where we explicitly state clustering which of course is an unsupervised learning technique!

PR #369, opened 2019-04-05
Some proposed changes

florisvdh:
Hi,

It is with much pleasure that I have been reading through the first five chapters of your book. Congratulations to the authors for such a fine result!

I hope my proposed (small) changes are of use to you.

Some more things which I've noted may follow as issues later on.


02-spatial-data.Rmd
```diff
@@ -900,7 +900,7 @@ nlayers(r_brick)
 ```
 
 A `RasterStack` is similar to a `RasterBrick` in the sense that it consists also of multiple layers.
-However, in contrast to `RasterBrick`, `RasterStack` allows you to connect several raster objects stored in different files or multiply objects in memory.
```
Robinlovelace: Definite fix, thanks.

02-spatial-data.Rmd
```diff
@@ -929,7 +929,7 @@ r_stack
 
 Another difference is that the processing time for `RasterBrick` objects is usually shorter than for `RasterStack` objects.
 
-Decision on which `Raster*` class should be used depends mostly on a character of input data. 
+Decision on which `Raster*` class should be used depends mostly on the character of input data.
```
Robinlovelace: :+1:

03-attribute-operations.Rmd
```diff
@@ -311,7 +311,7 @@ class(world_agg2)
 ```
 
 As illustrated above, an object of class `sf` is returned this time.
-`world_agg2` which is a spatial object containing 6 polygons representing the columns of the world.
+`world_agg2` which is a spatial object containing 6 polygons representing the continents of the world.
```
Robinlovelace: How did we miss that?

03-attribute-operations.Rmd
```diff
@@ -505,7 +505,7 @@ world %>%
   mutate(pop_dens = pop / area_km2)
 ```
 
-The difference between `mutate()` and `transmute()` is that the latter skips all other existing columns (except for the sticky geometry column):
+The difference between `mutate()` and `transmute()` is that the latter drops all other existing columns (except for the sticky geometry column):
```
Robinlovelace: Yes better wording.

03-attribute-operations.Rmd
```diff
@@ -620,7 +620,7 @@ levels(grain)
 ```
 
 This behavior demonstrates that raster cells can only possess one value, an identifier which can be used to look up the attributes in the corresponding attribute table (stored in a slot named `attributes`).
-This is illustrated in command below, which returns the grain size and wetness of cell IDs 1, 11 and 35:
+This is illustrated by the command below, which returns the grain size and wetness of cell IDs 1, 11 and 35:
```
Robinlovelace: Subtle but yes, this is an improvement.

03-attribute-operations.Rmd
```diff
@@ -641,7 +641,7 @@ Raster subsetting is done with the base R operator `[`, which accepts a variety
 
 Here, we only show the first two options since these can be considered non-spatial operations.
 If we need a spatial object to subset another or the output is a spatial object, we refer to this as spatial subsetting.
-Therefore, the latter two options will be shown in the next chapter (see Section \@ref(spatial-raster-subsetting) in the next chapter).
+Therefore, the latter two options will be shown in the next chapter (see Section \@ref(spatial-raster-subsetting)).
```
Robinlovelace: True - missed that duplication. Many thanks.

05-geometry-operations.Rmd
```diff
@@ -777,7 +779,7 @@ srtm_masked = mask(srtm, zion)
 ```
 
 Changing the settings of `mask()` yields different results.
-Setting `maskvalue = 0`, for example, will set all pixels outside the national park to 0.
+Setting `updatevalue = 0`, for example, will set all pixels outside the national park to 0.
```
Robinlovelace: Is this right @jannes-m ?

05-geometry-operations.Rmd
```diff
@@ -1107,7 +1109,7 @@ In R, vectorization refers to the possibility of replacing `for`-loops and alike
 
 The simplest form of vectorization is to convert the centroids of raster cells into points.
 `rasterToPoints()` does exactly this for all non-`NA` raster grid cells (Figure \@ref(fig:raster-vectorization1)).
-Setting the `spatial` parameter to `TRUE` ensures the output is a spatial object, not a matrix.
+Setting the `spatial` parameter to `TRUE` is needed to get a spatial object instead of a matrix.
```
Robinlovelace: :+1:

05-geometry-operations.Rmd
Nowosad: @Robinlovelace - that's correct

Nowosad:
@florisvdh thank you very much!

PR #380, opened 2019-05-24
Update 05-geometry-operations.Rmd

cshancock:
Nitpicky correction.


cshancock:
There are 49 political entities, but only 48 of them are states. The District of Columbia is an administrative division, and does not have the same political rights that the other states have. @Nowosad's proposed correction would be a good solution.

Robinlovelace:
Can you update the PR accordingly @cshancock ?

Robinlovelace:
Closing as this has gone stale. Feel free to re-open @cshancock.

PR #383, opened 2019-05-26
Remove triple colon...

prosoitos:
...which I think was a typo. Triple colon does exist, but I don't see why it would be necessary here. Apologies if I am wrong.

### DIFF
--- a/03-attribute-operations.Rmd
+++ b/03-attribute-operations.Rmd
@@ -301,7 +301,7 @@ class(world_agg1)
 
 The result is a non-spatial data frame with six rows, one per continent, and two columns reporting the name and population of each continent (see Table \@ref(tab:continents) with results for the top 3 most populous continents).
 
-`aggregate()` is a generic function which means that it behaves differently depending on its inputs. **sf** provides the method `aggregate.sf()` which is activated automatically when a `by` argument is provided, rather than using `~` to refer to the grouping variable:
+`aggregate()` is a generic function which means that it behaves differently depending on its inputs. **sf** provides the method `aggregate.sf()` which is activated automatically when `x` is an `sf` object and a `by` argument is provided:
 
 ```{r 03-attribute-operations-27}
 world_agg2 = aggregate(world["pop"], by = list(world$continent),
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Something like this (make sense?):

```
**sf** provides a method for the generic `aggregate` function. 
It is activated when an `sf` object is passed to the first `x` argument and the second `by` argument is a list of grouping variables:^[
The `~` notation does not result in `sf` outputs, e.g. with `aggregate(pop ~ continent, sum, data = world)`.
]
```

Robinlovelace:
For more on generics see https://adv-r.hadley.nz/s3.html#s3-methods

prosoitos:
> It was deliberate, but there was a typo in there. The true function is sf:::aggregate.sf().
> Although it is exported, it does not appear because it is a generic method.

Oh, I see. Thank you for the explanations. The package here is (to me at least) somewhat confusing since calling `?sf::aggregate` opens the help file for `aggregate.sf()`, but trying to run, for instance:

```r
world_agg2 = sf::aggregate(world["pop"], by = list(world$continent), FUN = sum, na.rm = TRUE)
```

leads to the `Error: 'aggregate' is not an exported object from 'namespace:sf'` message.

While running:

```r
world_agg2 = sf:::aggregate.sf(world["pop"], by = list(world$continent), FUN = sum, na.rm = TRUE)
```

works, but `?sf:::aggregate.sf` does not lead to the help file.

Seems that I better read Hadley's chapter on generic methods (thank you for the link).

As far as the wording goes, I actually find the initial wording quite clear (once the typo is corrected to `sf:::aggregate.sf`). How would you feel about this:

> `aggregate()` is a generic function which means that it behaves differently depending on its inputs. **sf** provides the method `aggregate.sf` which is activated automatically when a `by` argument is provided, rather than using the `~` to refer to the grouping variable:`

prosoitos:
I pushed a new commit with a wording suggestion. Let me know whether this seems clear.

Thanks!

Robinlovelace:
Looks good. Only change I'd recommend based on the latest version is mentioning that `x` must be an `sf` object. How would this be:

```
aggregate()` is a generic function which means that it behaves differently depending on its inputs. **sf** provides the method `aggregate.sf()` which is activated automatically when `x` is an `sf` object and a `by` argument is provided:
````

I think mentioning `~` could confuse people, so removed in this suggestion. Thanks for raising this.

prosoitos:
Looks good to me. I will push the change.

prosoitos:
I also added a link to Hadley's chapter on generics as I personally found it useful to understand this. But I am happy to rebase this last commit if you don't like the link.

Nowosad:
Nice changes! Thanks.

PR #387, opened 2019-05-27
Fix #386

### LINKED ISSUE #386: Errors in latest docker image
I've just updated the docker image so the build are fixed :tada:  

Unfortunately the book does not build out of the box, with these errors

- [ ] **bindrcpp** is not installed, creating an error message (easy to fix, but strange error seeing as we do not use the package)
- [ ] **spDataLarge** related error message (see below)

```Makefile:2: recipe for target 'html' failed
Quitting from lines 397-400 (13-location.Rmd) 
Error in FUN(X[[i]], ...) : 
  lazy-load database '/usr/local/lib/R/site-library/spDataLarge/data/Rdata.rdb' is corrupt
Calls: local ... library -> checkConflicts -> same.isFn -> vapply -> FUN
In addition: Warning message:
In FUN(X[[i]], ...) : internal error -3 in R_decompress1
Execution halted
Error in Rscript_render(f, render_args, render_meta) : 
  Failed to compile 13-location.Rmd
Calls: <Anonymous> ... render_new_session -> tryCatch -> tryCatchList -> Rscript_render
Execution halted
make: *** [html] Error 1
Error in render_book_script(output_format, envir, quiet) : 
  Error 2 attempting to render book
Calls: <Anonymous> -> <Anonymous> -> in_dir -> render_book_script
Execution halted

Exited with status 1.
```

Robinlovelace:


Robinlovelace:
Heads-up @jannes-m after these changes I can build the book in the latest docker image: 

image

Robinlovelace:
Yes that would do it. But assuming the reader has loaded the library, which is likely given https://github.com/Robinlovelace/geocompr/blob/a5a3fdd1d770855a90323bc69ae6dcaa5cedb1e7/13-location.Rmd#L13

13-location.Rmd
```diff
@@ -65,9 +65,9 @@ Although we have applied these steps to a specific case study, they could be gen
 ## Tidy the input data
 
 The German government provides gridded census data at either 1 km or 100 m resolution.
-The following code chunk downloads, unzips and reads in the 1 km data.^[
-`census_de` is also available from the **spDataLarge** package.
-]
+The following code chunk downloads, unzips and reads in the 1 km data.
+Please note that `census_de` is also available from the **spDataLarge** package (`data("census_de", package = "spDataLarge"`).
```
Robinlovelace: :+1:

Robinlovelace:
> However, I am not sure if every reader is aware of the fact, that datasets are attached invisibly by just calling the name of the dataset when the package they are stored in has been attached before.

May be worth saying that the data is 'lazily loaded' when the package is attached. Merging for now, thanks for the change, and agree, good to say that the objects are in the package in the main text.

PR #388, opened 2019-05-28
Minor correction in equivalence of read_sf() and st_read()

prosoitos:
Another difference of read_sf() is that it produces a tibble

### DIFF
--- a/07-read-write-plot.Rmd
+++ b/07-read-write-plot.Rmd
@@ -411,14 +411,14 @@ Instead of columns describing xy-coordinates, a single column can also contain t
 Well-known text (WKT), well-known binary (WKB), and the GeoJSON formats are examples of this.
 For instance, the `world_wkt.csv` file has a column named `WKT` representing polygons of the world's countries.
 We will again use the `options` parameter to indicate this.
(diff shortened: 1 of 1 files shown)


07-read-write-plot.Rmd
```diff
@@ -411,14 +411,14 @@ Instead of columns describing xy-coordinates, a single column can also contain t
 Well-known text (WKT), well-known binary (WKB), and the GeoJSON formats are examples of this.
 For instance, the `world_wkt.csv` file has a column named `WKT` representing polygons of the world's countries.
 We will again use the `options` parameter to indicate this.
-Here, we will use `read_sf()` which does exactly the same as `st_read()` except it does not print the driver name to the console and stores strings as characters instead of factors.
+Here, we will use `read_sf()`, the tidyverse-flavoured version of `st_read()`: strings are parsed as characters instead of factors and the resulting data frame is a [tibble](https://r4ds.had.co.nz/tibbles.html). The driver name is also not printed to the console.
```
Robinlovelace: The fact that it parses text strings into character vectors and not factors is an important difference. I knew that once but either forgot it or forgot to mention that important difference here. Many thanks for this very useful and for some people (tidyverse users, seemingly everyone now!) important change.

PR #392, opened 2019-06-03
Correct definition of qtm acronym

prosoitos:


prosoitos:
> Documents: https://rdrr.io/cran/tmap/man/qtm.html

There is also a section in the `vignette("tmap-getstarted")` (that's where I found the info).

PR #393, opened 2019-06-04
chap 9: suggest lower case and commas in the lists to fix sentence an…

prosoitos:
…d case incongruity

### DIFF
--- a/09-gis.Rmd
+++ b/09-gis.Rmd
@@ -51,23 +51,23 @@ The 'CLI vs GUI' debate can be adversial but it does not have to be; both option
 The advantages of a good CLI such as that provided by R (and enhanced by IDEs such as RStudio) are numerous.
 A good CLI
 
-- Facilitates the automation of repetitive tasks. 
-- Enables transparency and reproducibility, the backbone of good scientific practice and data science.
-- Encourages software development by providing tools to modify existing functions and implement new ones.
-- Helps develop future-proof programming skills which are in high demand in many disciplines and industries.
-- Is user-friendly and fast, allowing an efficient workflow.
+- facilitates the automation of repetitive tasks,
+- enables transparency and reproducibility, the backbone of good scientific practice and data science,
+- encourages software development by providing tools to modify existing functions and implement new ones,
+- helps develop future-proof programming skills which are in high demand in many disciplines and industries,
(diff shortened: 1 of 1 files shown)


09-gis.Rmd
```diff
@@ -36,7 +36,7 @@ of flexibility and power waiting for you with the command line. Many times
 you can do something on the command line in a fraction of the time you
 can do it with a GUI.
 
-The 'CLI vs GUI' debate can be adversial but it does not have to be; both options can be used interchangeably, depending on the task at hand and the user's skillset.^[GRASS GIS and PostGIS are popular in academia and industry and can be seen as products which buck this trend as they are built around the command-line.
+The 'CLI vs GUI' debate can be adversial but it does not have to be: both options can be used interchangeably, depending on the task at hand and the user's skillset.^[GRASS GIS and PostGIS are popular in academia and industry and can be seen as products which buck this trend as they are built around the command-line.
```
jannes-m: the copyediting suggested to change from colon to semicolon (see: https://github.com/Robinlovelace/geocompr/commit/b13212daebabf870943f0d57375aabf3318f2a24), so I would stick with the semicolon

Robinlovelace:
Good point @jannes-m. Both are fine but, in hindsight, minimising changes is a good thing, especially when they go against professional proof-readers (even though we may not always like their style!).

jannes-m:
Thanks for the input @prosoitos, we really appreciate your efforts. I won't merge the PR because I can't cherrypick, hence I will incorporate your suggestion to use lower case in these specific bullet lists manually.

prosoitos:
> minimising changes is a good thing

Sorry: I did suggest a lot of changes. I'll stop suggesting so many tiny edits!

> the copyediting suggested to change from colon to semicolon

I totally disagree with this change. Reading forums, grammar references, wikipedia, etc., etc. only conforts me in my belief that a colon is what is appropriate here. The 2nd sentence element explains the 1st (at least, this is how I understand the sentence). And this is a typical case for a colon.

But fair enough! :slightly_smiling_face: 

> I won't merge the PR because I can't cherrypick, hence I will incorporate your suggestion to use lower case in these specific bullet lists manually.

I would be more than happy to change that semi-colon on my end to save you from having to do anything manually.

PR #370, opened 2019-04-06
A few more small changes (chapters 5 & 6)

florisvdh:

PR #400, opened 2019-06-11
Few updates/fixes of chapters 7 and 9

florisvdh:
For some odd reason, it seems that GitHub still added commits here (chapters 5 & 6) that were already merged on April 9th (#370). Those commit hashes are different indeed from mine, but files 5 & 6 are not changed by this PR anyway.


09-gis.Rmd
```diff
@@ -100,6 +100,7 @@ Though not covered here, it is worth being aware of the interface to ArcGIS, a p
 The so-called R-ArcGIS bridge (see https://github.com/R-ArcGIS/r-bridge) allows R to be used from within ArcGIS. 
 One can also use R scripts from within QGIS (see https://docs.qgis.org/2.18/en/docs/training_manual/processing/r_intro.html).
 Finally, it is also possible to use R from the GRASS GIS command line (see https://grasswiki.osgeo.org/wiki/R_statistics/rgrass7).
+]
```
jannes-m: Yes, thanks! I am wondering when the closing square bracket disappeared....

09-gis.Rmd
```diff
@@ -513,7 +514,7 @@ link2GI::linkGRASS7(london_streets, ver_select = TRUE)
 
 Before we can use GRASS geoalgorithms, we need to add data to GRASS's spatial database.
 Luckily, the convenience function `writeVECT()` does this for us.
-(Use `writeRast()` in the case of raster data.)
+(Use `writeRAST()` in the case of raster data.)
```
jannes-m: Yes, you are absolutely right, thanks for spotting!

09-gis.Rmd
```diff
@@ -461,7 +462,7 @@ library(link2GI)
 link = findGRASS() 
 ```
 
-`link` is a `data.frame` which contains in its rows the GRASS 7 installations on your computer. 
+`link` is a `data.frame` which contains in its rows the GRASS installations on your computer.
```
jannes-m: Ok, I've just looked at the source code of `link2GI::findGRASS()`, and in fact, it looks for `grass??` which very likely means that also other older or newer GRASS executables would be found. Thanks for spotting!

jannes-m:
Thanks!

PR #414, opened 2019-06-13
Fix #411

### LINKED ISSUE #411: Requested clarification in section 7.7.2: raster file format
> The `raster` file format (native to the `raster` package) is used when a file extension is invalid or missing. 
Some raster file formats come with additional options.
You can use them with the `options` [parameter](http://www.gdal.org/formats_list.html).

The second sentence causes some confusion with the first, i.e. '`raster` file format' vs. 'raster file format' -- especially because they appear next to each other, within the same paragraph (a little lack of attention causes the reader to think it is still about the `raster` file format).

One solution may be to split it into separate paragraphs.

Robinlovelace:


Robinlovelace:
Good point @florisvdh. I've gave it some more thought and decided that `.grd` files should be mentioned early, before `.tif` files. See here for my additional changes: https://github.com/Robinlovelace/geocompr/commit/356671f5e76fb1e7532e5bb8cf50a899b502fee8

Does that seem OK?

florisvdh:
Yes, this seems a very good decision to me, thanks!

Issue #402, opened 2019-06-13
Requested clarification in sections 4.3.4/4.3.5: focal vs. zonal operations


florisvdh:
More clarity is needed on the distinction between the filters in both types of operation.

Currently, the following statements are a source of confusion:

- In the paragraph on focal operations:

> The neighborhood (also named kernel, filter or moving window) under consideration is typically of size 3-by-3 cells (that is the central cell and its eight surrounding neighbors), but can take on any other (not necessarily rectangular) shape as defined by the user.

- In the paragraph on zonal operations:

> *Zonal* operations are similar to focal operations.
> The difference is that zonal filters can take on any shape instead of a predefined rectangular window.

Perhaps (as I understand it), the primary distinction is that the focal operations use a 'moving window' while the zonal operations don't: these just apply an aggregation function per subgroup of raster cells as defined by another raster. Rather than the distinction being in the form of the filter.

jannes-m:
The distinction is that zonal operations use filters whose parts (raster cells) don't need to be neighbors, think for instance of land use categories or the example we have used in the book, soil grain sizes (sand, silt, clay; https://geocompr.robinlovelace.net/attr.html#fig:cont-raster). By comparison, the raster cells used as a focal filter have to be connected. 
If you can think of a better explanation in the book, please let us know.

florisvdh:
@jannes-m  Your answer already provides more clarity. The current confusion arises because it is not actually explained what 'zonal filters' are. From the two cited sentences in 4.3.5 (i.e. the first two of the zonal subsection), no clear distinction can be made compared to focal operations (where 'any other shape' is also mentioned). Then the text moves to an example.

I think a second distinction that could be pointed at more explicitly is the way in which zonal and focal filters are defined: the former are defined by a second raster's values, while the latter are defined explicitly in an abstract way (matrix of weights).

jannes-m:
Reading the two sentences instead of just the one you have cited and looking at the indicated figure, I think the difference is already made clear:

>Zonal operations are similar to focal operations. The difference is that zonal filters can take on any shape instead of a predefined rectangular window. Our grain size raster is a good example (Figure 3.2) because the different grain sizes are spread in an irregular fashion throughout the raster.
 
But if you have a better suggestion, please let us know!

florisvdh:
[...]

PR #416, opened 2019-06-14
Explaining more on zonal vs. focal

florisvdh:
This is an attempt to further clarify the differences between zonal and focal operations to the beginner, see #402 . But be critical and do make further corrections!

@jannes-m I saw that you already extended a bit in f803f5f, so you can compare and choose what to do.

### DIFF
--- a/04-spatial-operations.Rmd
+++ b/04-spatial-operations.Rmd
@@ -803,19 +803,16 @@ Chapter \@ref(gis) shows how to access such GIS functionality from within R.
 
 ### Zonal operations
 
-Just like focal operations, *zonal* operations aggregate values of multiple raster cells.
-Several aspects distinguish zonal from focal operations:
-
-- while the spatial filter in `focal()` is defined as a fixed matrix (to represent the moving window), 
-the *zonal filters* (or 'zones'), used to aggregate cell values, are areas defined by the unique values of a *second raster*.
-- consequently, the raster cells defining the zonal filter do not necessarily have to be neighbours.
-Hence zonal filters can take on many more shapes than spatial filters, which always consist of neighbouring cells (in a rectangular or other configuration).
+Just like focal operations, *zonal* operations apply an aggregation function to multiple raster cells.
+<!-- Several aspects distinguish zonal from focal operations:
+- A (rectangular) window of neighboring raster cells (e.g., 3x3 cells) usually defines the spatial filter of focal operations.  -->
+However, a second raster, usually a categorical raster, defines the *zonal filters* (or 'zones') in the case of zonal operations as opposed to a predefined neighborhood window in the case of focal operations (see previous Section).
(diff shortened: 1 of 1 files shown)


jannes-m:
Thanks for the suggestion and the effort. I have slightly rewritten your draft, let me know if you like it.

04-spatial-operations.Rmd
```diff
@@ -803,20 +803,24 @@ Chapter \@ref(gis) shows how to access such GIS functionality from within R.
...
+Just like focal operations, *zonal* operations apply an aggregation function to multiple raster cells.
+<!-- Several aspects distinguish zonal from focal operations:
+- A (rectangular) window of neighboring raster cells (e.g., 3x3 cells) usually defines the spatial filter of focal operations.  -->
+However, a second raster, usually a categorical raster, defines the *zonal filters* (or 'zones') in the case of zonal operations as opposed to a predefined neighborhood window in the case of focal operations (see previous Section).
+Consequently, the raster cells defining the zonal filter do not necessarily have to be neighbors.
+<!--Hence zonal filters can take on many more shapes than spatial filters, which always consist of neighbouring cells (in a rectangular or other configuration).-->
 Our grain size raster is a good example (right panel of Figure \@ref(fig:cont-raster)) because the different grain sizes are spread in an irregular fashion throughout the raster.
+Finally, the result of a zonal operation is a summary table grouped by zone which is why this operation is also known as *zonal statistics* in the GIS world.
```
florisvdh:
```suggestion
Finally, the result of a zonal operation is a summary table grouped by zone which is why this operation is also known as *zonal statistics* in the GIS world. 
By contrast, focal operations return a raster.
```

Robinlovelace: Looks good to me. Also, first time I've seen 'suggestions' in GitHub. That is awesome!

florisvdh:
This seems OK to me @jannes-m, thanks! The main distinction is very clear now.

I would suggest to be a little more explicit about the difference in output between focal and zonal, see my suggestion above. You can rewrite of course.

Robinlovelace:
I'm happy with this suggestion but will let @jannes-m enjoy the 'Commit suggestion' bit, assuming all good.

04-spatial-operations.Rmd
Robinlovelace: Hi @florisvdh I do not have permission to commit your suggestion. But please implement it and we can accept this PR I think. Heads-up @jannes-m .

florisvdh: Done. Seems like a Github bug to me though, as 'edits from maintainers' is marked for my branch.

Robinlovelace:
Thanks @florisvdh, this looks good to go!

jannes-m:
Thanks!

PR #424, opened 2019-08-04
'calcHistMatch()' was moved to 'satelliteTools'

patperu:
Little correction.

### DIFF
--- a/04-spatial-operations.Rmd
+++ b/04-spatial-operations.Rmd
@@ -875,7 +875,7 @@ For instance, we could compute the mean value.
 This might smooth the clear border in the merged result but it will most likely not make it disappear.
 To do so, we need a more advanced approach. 
 Remote sensing scientists frequently apply histogram matching or use regression techniques to align the values of the first image with those of the second image.
-The packages **landsat** (`histmatch()`, `relnorm()`, `PIF()`), **satellite** (`calcHistMatch()`) and **RStoolbox** (`histMatch()`, `pifMatch()`) provide the corresponding functions.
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Thanks for this @patperu

Robinlovelace:
Sounds like a good plan Jannes.

patperu:
Oh yes, good point! So, you will find a solution, right?
BTW I found this function because I was looking for a way to replicate this analysis [https://github.com/vis4/eu-drought](https://github.com/vis4/eu-drought) with R.

Robinlovelace:
Worth contacting the package developers to see what they recommend?

Robinlovelace:
Any update on this @patperu ? Looking at it it seems the package hasn't changed but I don't want the book to be out-of-date.

Robinlovelace:
What do you think @Nowosad and @jannes-m ?

jannes-m:
The correction is correct. Unfortunately, package **satelliteTools** is only available on [github](https://github.com/environmentalinformatics-marburg/satelliteTools/tree/master/R). So there are three options:

1. say in the text, that **satelliteTools** is only available on github
1. delete the reference to **satelliteTools**
1. say in a footnote that there is another option available in a github package

Robinlovelace:
Thanks @jannes-m, I think all 3 options sound viable. This sounds the simplest:

> say in the text, that satelliteTools is only available on github

Robinlovelace:
Superseded by https://github.com/Robinlovelace/geocompr/commit/f918610935350a726e1d71f7f36c0f4eab92dd88

PR #426, opened 2019-08-21
update Dockerfile with QGIS3 installation instructions

jannes-m:
Hey Robin,
I have just added the stuff to the Dockerfile that will also install the latest QGIS LTR release. This way we will be able to run also RQGIS3, rgrass7 and RSAGA from within our geocompr docker container.
Regarding your dream to open up the the docker container inside the geocompr project folder, I suggest the following:
  
```sh
cd ~/
# the gecompr repository into your home 
# git clone git@github.com:Robinlovelace/geocompr.git 
git clone https://github.com/Robinlovelace/geocompr.git 
# run rocker-docker geocompr docker container while mounting local geocompr folder
cd geocompr
# then start rocker docker geocompr
docker run -e PASSWORD=pass -p 8787:8787 -v $(pwd):/home/rstudio/geocompr robinlovelace/geocompr
```
Now clicking the .Rproj file will also make available the git pane in RStudio. Please note that working on the geocompr files will change them also on disk, not only within the docker container!

The following does not work as expected. It creates the folder `home/rstudio/geocompr` but it does not `cd` into it: `docker run -e PASSWORD=pass -p 8787:8787 -v $(pwd):/home/rstudio/geocompr/ --workdir="/home/rstudio/geocompr/" robinlovelace/geocompr`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,34 @@
 FROM rocker/geospatial
 RUN R -e "remotes::install_github('r-spatial/lwgeom')"
 RUN R -e "remotes::install_github('geocompr/geocompkg')"
+# install RQGIS3 from github
+RUN R -e "remotes::install_github('jannes-m/RQGIS3')"
+RUN apt-get update && \
+  # software-properties-common contains add-apt-repository needed for ubuntugis
+  # apt-get install -y gpg software-properties-common && \
+  apt-get install -y gpg && \
+  # install QGIS and dependencies
+  sh -c 'echo "deb http://qgis.org/debian-ltr stretch main" >> /etc/apt/sources.list' && \
+  sh -c 'echo "deb-src http://qgis.org/debian-ltr stretch main" >> /etc/apt/sources.list' && \
+  # installing ubuntugis-unstable did not work
+  # sh -c 'echo "deb http://qgis.org/ubuntugis-ltr xenial main" >> /etc/apt/sources.list' && \
+  # sh -c 'echo "deb-src http://qgis.org/ubuntugis-ltr xenial main" >> /etc/apt/sources.list' && \
+  # sh -c 'echo "deb http://ppa.launchpad.net/ubuntugis/ubuntugis-unstable/ubuntu xenial main" >> /etc/apt/sources.list' && \
+  # add-apt-repository ppa:ubuntugis/ubuntugis-unstable && \
+  wget -O - https://qgis.org/downloads/qgis-2019.gpg.key | gpg --no-tty --import && \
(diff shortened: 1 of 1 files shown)


jannes-m:
BTW the Dockerfile found on https://github.com/jannes-m/docker-rqgis/tree/master/rqgis_2.18 lets you still use the last QGIS 2.18 release, i.e., QGIS 2.18.26. Our book could be fully reproduced, including the R-GIS code chunks, using this docker container. However, one has to clone the old QGIS repo and compile it from source which takes a while.

Robinlovelace:
> Regarding your dream to open up the the docker container inside the geocompr project folder, I suggest the following:

Could that be done in the Dockerfile?

jannes-m:
well, if you add following two lines to the Dockerfile:

```sh
RUN git clone --progress https://github.com/Robinlovelace/geocompr /home/rstudio/geocompr
WORKDIR /home/rstudio/geocompr
```
And then run in the terminal: `sudo docker run -it robinlovelace/geocompr /bin/bash` you would see that the working directory in fact is `/home/rstudio/geocompr`. However, it does not work when doing the same using RStudio, but at least the geocompr folder would be accessible in RStudio. Still, I like the other approach - mounting a local folder to the docker container - better, at least for my usual workflow. But maybe when sharing the container it might be better to have the geocompr repo in the container, not sure and I don't mind, do as you like.

jannes-m:
Just pulled the latest geocompr container and it worked, QGIS3 etc. are now available!! :tada:

Robinlovelace:
Great to hear it's working @jannes-m, mission accomplished!

jannes-m:
Should I add the last two lines to the Dockerfile? Then we would also have the geocompr repo inside the docker container.

Robinlovelace:
I can see no downside, only a large upside. So yes, please do @jannes-m :+1:

Robinlovelace:
How about this: 

```
RUN git clone --progress https://github.com/Robinlovelace/geocompr /home/rstudio/geocompr & \
  cp -R /home/rstudio/geocompr/* /home/rstudio/
```

?

```

**Comment by jannes-m** [comment]
To achieve the same you would just have to clone the geocompr repo into the rstudio folder (which is easier), I guess like this:

```
RUN git clone --progress https://github.com/Robinlovelace/geocompr /home/rstudio
```
Yes, possible, however then there is this strange kitematic folder within our geocompr project which is not necessarily harmful just ugly :smile:

Robinlovelace:
Don't let unugly be the enemy of good :pig:  :smile:

PR #428, opened 2019-08-22
Allow packages to update with CRAN versions

Robinlovelace:
See https://github.com/rocker-org/rocker-versioned/issues/137

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ RUN apt-get update && \
   xvfb && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*  && \
+  echo "options(repos = c(CRAN = 'https://cran.rstudio.com/'), download.file.method = 'libcurl')" >> /usr/local/lib/R/etc/Rprofile.site \
   # install virtual display for Python
   pip3 install pyvirtualdisplay
   


jannes-m:
Just fyi, to update packages (not only R packages, also other software packages such as QGIS, etc.) within a docker container, I just update the packages inside the container and afterwards do `docker commit ...` to make the changes persistent.

Robinlovelace:
Thanks for the info, didn't know that, and potentially very useful. Does it also work when updating containers? Imagine so... https://docs.docker.com/engine/reference/commandline/commit/

Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN apt-get update && \
   xvfb && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*  && \
+  echo "options(repos = c(CRAN = 'https://cran.rstudio.com/'), download.file.method = 'libcurl')" >> /usr/local/lib/R/etc/Rprofile.site \
```
cboettig: yup, this looks good.  (though we do it as well, dictating `download.file.method` is arguably a bit aggressive, though libcurl should be available on these images and I believe will enable parallel downloads.

Robinlovelace: Thanks for the feedback @cboettig!

jannes-m:
Regarding `docker commit ...` `install.packages(...)` will work when running RStudiothe rocker docker container on a localhost, you have to add the ROOT option in order to run `sudo apt-get upate` from within the terminal, e.g.,

```docker
docker run -d -p 8787:8787 -e ROOT=TRUE -e PASSWORD=pass robinlovelace/geocompr 
```
After stopping the container, run `docker ps -a` to find out the container id, let's say it is a5c4b6d33946, and then you can run the commit command:

```docker
docker commit a5c4b6d33946 robinlovelace/geocompr
``` 
I don't know if you can then push the changes also to dockerhub, I haven't tried that nor looked it up. The disadvantage of the `commit` approach is, that it might hinder reproducibility, say you install new R packages or other software not listed in the Dockerfile. In these cases I would prefer to update the Dockerfile.

Robinlovelace:
Cheers Jannes, useful to know. BTW are you happy with this PR being merged?

Issue #429, opened 2019-08-22
Add binder?


Robinlovelace:
Experimenting with binder here: https://github.com/geocompr/RGSWorkshop

Let's see what comes of it - give it a try if you get a chance and let me know if it works!

Robinlovelace:
Heads-up: it works!

![image](https://user-images.githubusercontent.com/1825120/63516007-05c69a00-c4e4-11e9-803e-b0086284e916.png)

jannes-m:
@pat-s will use it at the summer school next week

Robinlovelace:
Worth adding a binder link to this repo? I suspect so.

Robinlovelace:
I think that PR request will close that issue :tada: 

Can you test it here and review the PR @jannes-m and @Nowosad ?

https://github.com/Robinlovelace/geocompr/tree/binder

Robinlovelace:
It works! 

![image](https://user-images.githubusercontent.com/1825120/63574735-61defc00-c580-11e9-973b-fce6953c7850.png)

PR #430, opened 2019-08-22
Binder

Robinlovelace:
#429


Robinlovelace:
Merging as small change, build is passing, and it needs to be on master for Binder to work.

PR #436, opened 2019-08-26
Update refs

Robinlovelace:


Robinlovelace:
Heads-up @jannes-m the git clone bit in Docker seems to be causing the tests to fail: https://cloud.docker.com/repository/docker/robinlovelace/geocompr/builds/4450bc57-b602-4ba1-8c12-8cd2fd9f3cd6

jannes-m:
Don't think it has to do with the latest commit since we changed only the directory path where to install our repo. I can't see any build error log files either on dockerhub so maybe the build just took to long, see https://success.docker.com/article/docker-hub-automated-build-fails-and-the-logs-are-missing-empty? But I can try to run it locally and report back. Will do so during the day when I have a fast internet connection.

jannes-m:
Ok, I have run it locally and it doesn't work either. I could get it to work but then we still face the .Rproj problem. I couldn't make this work. So, I propose we do not do the git clone command within the Dockerfile but say instead that one could mount the geocompr to the docker container if it the geocompr repo was cloned locally before (which is basically what you already show in our README). Ok?

Robinlovelace:
Or we download and unzip the files. Sure it's doable and worth it: it's really cool being able to rebuild the book in docker (and binder, tested).

jannes-m:
I would suggest the following procedure: let us delete the `git clone` command from the Dockerfile. Otherwise, we would also have to update the docker image each time we change our repo, not a good idea. What is more not all people want to start our docker image in combination with our geocompr repo. Those people could then just run:

```sh
docker run -p 8787:8787 -e DISABLE_AUTH=TRUE /robinlovelace/geocompr
```

All the people who want to have the geocompr repo available in the docker container can do instead:

```sh
cd /to/geocompr/repo
# update to the latest version
git pull
# mount the current working directory to the docker container
docker run -d -p 8787:8787 -e DISABLE_AUTH=TRUE --volume $(pwd):/home/rstudio/geocompr  geocompr_local
```
Then you can also open the .Rproj.

Robinlovelace:
> let us delete the git clone command from the Dockerfile.

Sounds reasonable @jannes-m, but if there is a way to get the content of the book into the Docker container, I don't see any big disadvantage. The docker container is a snapshot and shipping with the latest version of the repo is a good plan I think. Perhaps downloading the .zip file and giving it an appropriate owner, e.g. rstudio, could solve this issue ? 

I'm really keen on keeping the geocompr source code in the Docker container now I've seen it in Binder but happy for the clone line to be removed for now. In any case, I think we can merge this PR as it's unrelated.

jannes-m:
yeah, you can merge as it is really unrelated. But using the second option, the geocompr repo will be available in the Docker container (`docker run -d -p 8787:8787 -e DISABLE_AUTH=TRUE --volume $(pwd):/home/rstudio/geocompr  robinlovelace/geocompr`), hence I am not quite sure that I understand what you are referring to. But ok, I will just update the dockefile and then you can run it, and you'll see if you like it.

Robinlovelace:
> the geocompr repo will be available in the Docker container (`docker run -d -p 8787:8787 -e DISABLE_AUTH=TRUE --volume $(pwd):/home/rstudio/geocompr  robinlovelace/geocompr`)

For sure, that's what I usually do. I think there are advantages of it being in there by default. Wonder how Binder does it. May be worth taking a look, wonder if this does it:
https://github.com/Robinlovelace/geocompr/blob/7bc06895d29529fe4356a5137905e8f12bcf4886/.binder/Dockerfile#L4

jannes-m:
`COPY .  ${HOME}` just copies the contents of your local directory (in our case our local geocompr repo) into the home folder of the docker image. I can try this tomorrow when I have again a good internet connection. But what is the advantage to have the geocompr repo inside the docker container?

Robinlovelace:
It encourages people to play with the source code, which in turn will increase the learning opportunities provided by the geocompr book, and make it easier for people to see how fixes could improve the book, by being able to rebuild the entire book in every geocompr Docker image.

jannes-m:
ok, I tried the `COPY . /home/rstudio/geocompr`. It works, however, you can still not open the .Rproj file, so I suggest to just mount the local geocompr repo. In fact, I can't really see a big difference between having the repo inside the container or mounting it to it. But I will keep my eyes open so eventually we will stumble across the solution.

PR #445, opened 2019-09-10
Fix #444

### LINKED ISSUE #444: Install ImageMagick on Docker
See here - testing now in robinlovelace/geocompr

Robinlovelace:


Dockerfile
```diff
@@ -40,6 +38,10 @@ RUN apt-get update && \
   make && \
   make install && \ 
   ldconfig /usr/local/lib && \
+  apt-get clean && \
```
Robinlovelace: Great idea, tidy up in the right place. Thanks @jannes-m

Robinlovelace:
> Ok, looks good. The only thing we have to keep in mind is that our docker image is getting bigger and bigger. I guess this is the reason why you have added docker/mini/Dockerfile.

Exactly.

Magick is tiny compared with QGIS so relatively small difference. Will try to get the `min` tag building on Dockerhub. 

Potentially the most useful thing about this PR is that it provides access to the whole Python stack, thanks to your installation of Python 3 things. Many thanks for the review.

Issue #443, opened 2019-09-02
Change RStudio defaults in Docker + Binder


Robinlovelace:
I think we should make code in .Rmd files evaluate in the console.

See here for how to do that: https://github.com/rstudio/rstudio/issues/1607#event-2501270331

Reasoning: https://github.com/rstudio/rstudio/issues/5280

Robinlovelace:
Closing for now as we have updated the .Rmd setting for RStudio preview. Any other preferences to change? Worth opening as a new issue.

PR #447, opened 2019-09-19
Fix 443

Robinlovelace:
This only kicks in in daily builds of Rstudio, but the setting will work in a few months when RStudio 1.3 is released. See https://github.com/rstudio/rstudio/issues/5301#issuecomment-533319762


Robinlovelace:
Should fix #443, eventually.

PR #449, opened 2019-10-04
Include quotes in table of style tags

nickbearman:
When actually writing the `style` code you need quote marks in the attribute, so i.e. `style = cont` won't work, but `style = "cont"` will. I suggest adding the quote marks to the entries in the table.

### DIFF
--- a/08-mapping.Rmd
+++ b/08-mapping.Rmd
@@ -257,12 +257,12 @@ In addition to manually setting `breaks` **tmap** allows users to specify algori
 \index{tmap (package)!break styles}
 Here are six of the most useful break styles:
 
-- `style = pretty`, the default setting, rounds breaks into whole numbers where possible and spaces them evenly;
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Good point @nickbearman, many thanks for the input. Not sure why the build is failing, it may be that the quotes messed something up, but I'm sure your suggestion is a step in the right direction. Also this may affect the LaTeX version (cc @Nowosad ).

Nowosad:
Hi @nickbearman, thanks a lot. I will take a look at it during the weekend.

Nowosad:
I think the build error is related to https://github.com/rspatial/raster/issues/72. I think the best approach for now is to wait a few days...

Nowosad:
BTW @nickbearman can you add the quotes around `order` too?

nickbearman:
Great stuff - added in quotes around order

Robinlovelace:
Thanks for the detective work @Nowosad and the updated PR @nickbearman. I'd say this is good to go!

Nowosad:
I've checked it - works great. Thanks once again @nickbearman !

PR #454, opened 2019-11-28
Update 04-spatial-operations.Rmd

RobinKohrs:
Kind of unnecessary, but I think there is a missing "n"


Robinlovelace:
Many thanks @RobinKohrs :+1:

PR #462, opened 2020-01-13
No interactive maps

Robinlovelace:


Robinlovelace:
Assuming this branch passes Travis tests I tentatively suggest we merge this on. One to discuss @Nowosad what are your thoughts?

Nowosad:
Hi @Robinlovelace - I would suggest waiting a few days before merging any PR. I will look at the changes before the end of the week and give you my feedback..

Robinlovelace:
OK great. All of this is in the context of the changes happening to packages upstream. For now I would favor reliability over visuals. According to https://github.com/rstudio/bookdown/issues/15#issuecomment-573776659

> leaflet::addProviderTiles() cannot work with the K-M method of bookdown. 

I tried the `M-K` approach but think that is even more problematic, hence trying to purge the book of interactive maps. I think if we only use maps that do not use `leaflet::addProviderTiles()` that may solve the problem but have not tried it. 

Looking forward to hearing the results of your investigations and hoping we can get the build passing on travis by the end of the week.

andrewmaclachlan:
@Robinlovelace @Nowosad  did you manage to find a solution, i'm getting the same error as you now. M-K approach also doesn't work, do understand what has happened, unless there has been an update that has affected leaflet

Robinlovelace:
@andrewmaclachlan take a look here: https://github.com/Robinlovelace/geocompr/pull/465/files

andrewmaclachlan:
Thanks @Robinlovelace , looks to be back up and running now! Did you raise an issue with leaflet?

Robinlovelace:
No because it seems that others, including people who work for RStudio (which funds development of bookdown and leaflet), have identified the issue so it's known but seems to be considered as a 'wontfix'.

PR #489, opened 2020-04-05
Rqgis3

Robinlovelace:


Robinlovelace:
Heads-up @jannes-m and @pat-s I've opened this PR to track changes for QGIS3 in geocompr. Sorry for nagging but any idea of when this will be ready? I suggest May could be a good time to revive work on this when changes in the R and QGIS world have settled down.

pat-s:
Unfortunately I cannot promise any ETA here or at all.
Besides starting a new job in May in a new country, RQGIS3 is not among my priorities right now.

In addition, I haven't had any contact to Jannes in the last months and he also does not seem to have time for open-source work right now (seen on his commit history).
I was pushing on archiving RQGIS for the last months but we still haven't managed to do so yet.

I guess it would make sense for someone else to take over but I fear there might be nobody available. However, in the end it's Jannes project and he will always have the last word on this - I am just a collab :) Maybe he still has plans to continue all of this soon (I honestly don't know).

So in summary: Things are not looking good right now for RQGIS3 in the near future.

Robinlovelace:
Thanks for the quick response @pat-s and congrats on the new job! I will communicate with Jannes to decide a best way forward. If RQGIS is archived and RQGIS3 does not make it to CRAN that would be a great pity because they are very useful packages, and would also require substantial re-write of the GIS bridges chapter...

jannes-m:
Hey  Patrick, I am really sorry, I really didn't want to ignore you! But with the lockdown I have even less time. My priorities right now are: changing diapers, consoling crying kids, carrying them around, surviving the dreadful nights and making it work with Kat. In between I try to work a few hours at home but I hardly have time to fire up my private laptop and if I can I only have a few minutes... So in short, I am really sorry but right now I can't maintain and develop RQGIS(3) especially when considering the many pitfalls (RStudio bug, constant changes to API, differences between OS). So I have no objections to anyone taking over though I don't know who that would be but let's see. And yes, I even googled for a few minutes how to archive RQGIS but didn't find a solution quickly which is why I haven't done so so far... I have also played around with the mlr3 spatio-temporal stuff but yes, I need to get back to you via another channel, hopefully this week to agree on how to further proceed. 

Therefore, it might be a good idea to leave the RQGIS stuff out of the book for now but let's talk about this in person I would suggest.

Robinlovelace:
Thanks @jannes-m, no need for apologies, circumstances always change and family should definitely come first. One idea: I could do a 'call out' to see if anyone wants to help maintain this... I plan to ask rOpenSci if they can support work on RQGIS3. Sound reasonable. Happy to talk on the phone also. Closing this for now.

PR #491, opened 2020-04-10
Ghactions

Nowosad:


Robinlovelace:
Great work @Nowosad let me know if you want me to take a look at any point.

Nowosad:
Will do - I am getting closer to the solution.

Nowosad:
@Robinlovelace see https://github.com/Robinlovelace/geocompr/runs/576951334?check_suite_focus=true

Robinlovelace:
I think this is good to go, Netlify auto-deploys from the `gh-pages` branch so I don't think we need that action, the only one failing by the looks of it.

Nowosad:
So feel free to merge. We can deal with the second workflow later.

PR #528, opened 2020-04-23
improves cache

Nowosad:
I think the cache has not been working properly (about 30 minutes for packages installation each time). It should fix it.


Nowosad:
It should be good now. GH Actions time decreased from about 30 minutes to less than 8 minutes. :fireworks:

Issue #543, opened 2020-06-13
Build is failing


Robinlovelace:
With the following [error message](https://github.com/Robinlovelace/geocompr/runs/765925943#step:10:3598)

```
Quitting from lines 1170-1191 (05-geometry-operations.Rmd) 
##[error]Error in st_crs.character(x) : invalid crs: unknown
Calls: local ... pretty30 -> <Anonymous> -> st_crs -> st_crs.character
```

Robinlovelace:
I cannot reproduce this on my computer with the following set-up:

``` r
library(sf)
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 7.0.0
```

<sup>Created on 2020-06-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 3.6.3 (2020-02-29)
#>  os       Ubuntu 18.04.4 LTS          
#>  system   x86_64, linux-gnu           
#>  ui       X11                         
#>  language en_GB:en                    
#>  collate  en_GB.UTF-8                 
#>  ctype    en_GB.UTF-8                 
#>  tz       Europe/London               
#>  date     2020-06-13                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date       lib source                             
#>  assertthat    0.2.1      2019-03-21 [2] CRAN (R 3.6.0)                     
#>  backports     1.1.7      2020-05-13 [1] CRAN (R 3.6.3)                     
#>  callr         3.4.3      2020-03-28 [1] CRAN (R 3.6.3)                     
#>  class         7.3-17     2020-04-26 [2] CRAN (R 3.6.3)                     
#>  classInt      0.4-3      2020-04-06 [1] Github (r-spatial/classInt@d024051)
#>  cli           2.0.2      2020-02-28 [1] CRAN (R 3.6.2)                     
#>  crayon        1.3.4      2017-09-16 [2] standard (@1.3.4)                  
#>  DBI           1.1.0      2019-12-15 [2] CRAN (R 3.6.2)                     
#>  desc          1.2.0      2018-05-01 [2] standard (@1.2.0)                  
#>  devtools      2.3.0      2020-04-10 [1] CRAN (R 3.6.3)                     
[...]

Issue #519, opened 2020-04-20
Update GIS bridges chapter


Robinlovelace:
- [x] No longer depend on RQGIS for book build 
- [x] Purge RQGIS code from chapter
- [x] ~~Explain how to use `qgis_process` from system~~
- [ ] rgee section 
- [ ] `sf::gdal_utils()`
- [ ] More on `gdalUtils` - e.g. with different versions via Docker?
- [x] Something on Whitebox
- [x] ~~Something on Reticulate?~~

https://cran.r-project.org/web/packages/RQGIS/index.html

I've just tried to build the book without RQGIS and only 2 chapters are affects. Looking at fixes now...

Robinlovelace:
Looking for a hotfix in https://github.com/Robinlovelace/geocompr/pull/520

Robinlovelace:
Update: this is harder than initially anticipated and I'm not sure #520 is the best approach...

Robinlovelace:
Update: triggering build after https://github.com/geocompr/geocompkg/commit/57e159e059190943d7fb447edd81fb9d8909c079 - hoping that will at least fix the build. Wider issue is how to provide futureproof info on RQGIS. I suggest asking for support with RQGIS3 on the r-spatial discuss repo and in the wider community.

jannes-m:
I think the easiest solution would be to use `devtools::install_version("RQGIS", version = "1.0.4") `. What also should work is `
install.packages("https://cran.r-project.org/src/contrib/Archive/RQGIS/RQGIS_1.0.4.tar.gz", repos = NULL, type = "source")`

jannes-m:
With regard to RQGIS3 we can of course ask the community for help. Another solution is to simply drop the RQGIS part from the GIS chapter.

Robinlovelace:
> With regard to RQGIS3 we can of course ask the community for help. 

How about posting something like this on the `r-spatial/discuss` issue tracker:

Is anyone interested in developing RQGIS3? The original package RQGIS has been archived due to updates to QGIS and the original creators of the package lack the time, at present, to take the package further on it's journey towards CRAN.

jannes-m:
And another suggestion: we use RQGIS in most chapters (aside from the GIS chapter) for attaching data shipped with the package. I suggest to move this data to spData.

jannes-m:
[...]

PR #549, opened 2020-06-13
Reduce deps fix build

Robinlovelace:


Robinlovelace:
See  #543

Robinlovelace:
Merging this now @Nowosad - a step in the right direction. Any comments post-merge welcome. Priority I think should be to remove dependency on QGIS. #519

PR #550, opened 2020-06-14
load purrr for set_names()

theFippo:
set_names() used in 3.2.4 probably refers to function from purrr and needs to be loaded first


theFippo:
My pleasure. It's a great book to learn from after all. I agree, loading purrr isn't really necessary as setNames does the job already.

PR #553, opened 2020-06-15
Test dev version of sf

Robinlovelace:


Robinlovelace:
Nudge @Nowosad can you take a look when you get a chance?

Nowosad:
Hi @Robinlovelace, can you quickly explain why do we need this change?

Robinlovelace:
The aim is to test the book's code on the dev version of `sf` so we can be aware of issues before `sf` 1.0.0, which will support S2, goes to CRAN. There are other ways we could do this including via Docker images and a dedicated `geocompr-tests` repo that I considered but I think this is the simplest way to test the development version of `sf` that I came up with. There may be other ways and I wanted to get your thoughts before we merge this.

Nowosad:
@Robinlovelace I have a feeling that Docker would be an easier option here. We can use an existing Docker image, and just run two lines of code - (1) for sf dev installation, and (2) to build the book. What do you think about it?

Robinlovelace:
> We can use an existing Docker image, and just run two lines of code - (1) for sf dev installation, and (2) to build the book. What do you think about it?

Good idea outlined in this link you sent: https://jozef.io/r922-github-actions-r-packages/

Another advantage of that approach: it could be quicker. Are you up for giving it a go @Nowosad ?

Nowosad:
@Robinlovelace I could, but not sooner than sometime next week.

Robinlovelace:
Thanks @Nowosad. Any downsides of merging this as a stop-gap before then?

PR #577, opened 2020-10-04
Fixgetdata

Robinlovelace:


Robinlovelace:
I think this is good to go... Will fix the docker builds so keen to get it in fairly soon, unless they fix it upstream within a day or so... Heads-up @Nowosad

Robinlovelace:
Merging for now - we can always re-open if/when `getData()` is fixed on CRAN.

Nowosad:
Thanks @Robinlovelace

PR #600, opened 2021-01-25
Fix typos in section 10.4: Functions

zmbc:
This PR fixes two small issues in the "Functions" section:
- The function's argument was given the name `x`, which both broke later references to `poly_mat` and shadowed an inner function's argument.
- The line defining `Origin` was missing.

### DIFF
--- a/10-algorithms.Rmd
+++ b/10-algorithms.Rmd
@@ -339,8 +339,9 @@ Note that the functions we created are called iteratively in `lapply()`\index{lo
 ]
 
 ```{r 10-algorithms-15}
-poly_centroid = function(x) {
-  i = 2:(nrow(x) - 2)
+poly_centroid = function(poly_mat) {
+  Origin = poly_mat[1, ] # create a point representing the origin
+  i = 2:(nrow(poly_mat) - 2)
   T_all = lapply(i, function(x) {
     rbind(Origin, poly_mat[x:(x + 1), ], Origin)
   })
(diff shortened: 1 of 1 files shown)


Robinlovelace (approved):
Thanks for this. I think it does increase clarity (but still succeeded before).

Looking at previous mentions of the object name in the chapter this is consistent: 

```
rg poly_mat                                
10-algorithms.Rmd
95:The work undertaken by this script is demonstrated in the reproducible example below, which works on a pre-requisite object named `poly_mat`, a square with sides 9 units in length (the meaning of this will become apparent in the next section):^[
101:poly_mat = cbind(
109:poly_mat = cbind(
169:poly_mat = cbind(x_coords, y_coords)
177:Origin = poly_mat[1, ]
179:T1 = rbind(Origin, poly_mat[2:3, ], Origin) 
187:plot(poly_mat)
188:lines(poly_mat)
228:i = 2:(nrow(poly_mat) - 2)
230:  rbind(Origin, poly_mat[x:(x + 1), ], Origin)
251:The great thing about *scripting* the algorithm is that it works on the new `poly_mat` object (see exercises below to verify these results with reference to `st_centroid()`):
260:if we aimed only to find the centroid\index{centroid} of a polygon, it would have been quicker to represent `poly_mat` as an **sf** object and use the pre-existing `sf::st_centroid()` function instead.
```

Robinlovelace:
Thanks for another great contribution @zmbc !

PR #603, opened 2021-02-03
Update 04-spatial-operations.Rmd

MikeJohnPage:
Added missing 'and'.

### DIFF
--- a/04-spatial-operations.Rmd
+++ b/04-spatial-operations.Rmd
@@ -30,7 +30,7 @@ These are covered in Section \@ref(topological-relations), on topological relati
 
 \index{spatial operations}
 Another unique aspect of spatial objects is distance.
-All spatial objects are related through space, and distance calculations can be used to explore the strength of this relationship. This is covered in Section \@ref(distance-relations).
+All spatial objects are related through space, and distance calculations can be used to explore the strength of this relationship. These are covered in Section \@ref(distance-relations).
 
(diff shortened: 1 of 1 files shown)


Robinlovelace (commented):
I think the original text is fine. Remove the clause in commas and it reads:

> All spatial objects are related through space and distance calculations can be used to explore the strength of this relationship.

Would the section be better as:

> All spatial objects are related through space, and distance calculations can be used to explore the strength of this relationship. These are covered in Section xx.

?

MikeJohnPage:
Ah, having read your second suggestion I understand where the misunderstanding has happened. I read the original quote like this (notice the added comma):

> All spatial objects are related through space and distance, calculations can be used to explore the strength of this relationship.

I just showed the original text to a peer and they made the same mistake. Perhaps your second suggestion is better in this case?

> All spatial objects are related through space, and distance calculations can be used to explore the strength of this relationship. These are covered in Section xx.

Robinlovelace:
> Perhaps your second suggestion is better in this case?

I agree. Could you update the Pull Request?

Thanks for flagging this, always room for improvement.

MikeJohnPage:
> Thanks for flagging this, always room for improvement.

No problem, thanks for writing a great resource

> Could you update the Pull Request?

Done 👍

Robinlovelace (approved):

PR #633, opened 2021-09-06
Update .gitignore

Robinlovelace:


Robinlovelace:
Minor change that will enable me to do `git add .` without adding the libs files.

PR #640, opened 2021-09-20
Update figure 1.2, close #639

### LINKED ISSUE #639: Update Figure 1.2 shown n. downloads of geographic packages
I think we should:

- [x] Remove ggmap
- [x] Add terra
- [x] Add stars
- [x] Make the figure wider than it is tall
- [ ] Anything else?

We could use a log scale for the Y axis. Thoughts? I'm up for giving this a go. Currently the figure is as follows:

image

Robinlovelace:


Robinlovelace:
Heads-up @Nowosad that latest commit results in this (note it's only about class system packages now):

image

We could save another plot for the vis chapter...

PR #658, opened 2021-10-10
Update text in c3 for v2

Robinlovelace:


Robinlovelace:
Minor changes added in spare moment between family commitments : )

PR #682, opened 2021-11-27
Replace mentions of raster with terra where appropriate

iod-ine:
I noticed two incorrect mentions of the `raster` package instead of `terra`. Both of these refer to the contents of the book, and I believe they are just typos. This pull request fixes them. I also checked for all mentions of `**raster**` in the first seven chapters using `grep` and read through the context to ensure the rest are correct.


Nowosad:
Great! Thank you, Ivan!

Robinlovelace:
This is great stuff. Many thanks @iod-ine from me too and any other issues you spot let us know!

PR #683, opened 2021-11-27
Modify an example in ch4 and add a solution

iod-ine:
The last exercise in chapter 4 does not have a solution at the moment. In e1c41fd6885ea3966539ef63e9e6168ce08c7414 @Nowosad commented it out with a to-do of improving/replacing/modifying it. I think it's a good exercise, so I added a solution to the first part, which asks to calculate distance to the coast for Spain.

I removed the second part that asks to calculate weighted distance, at least for now. I could not figure out how to complete this from the text of the chapter and the documentation of `terra::distance`. I am more than willing to explore it further if you can guide me in the right direction :)

I also include a reproducible example of the code I added as a solution.

``` r
library(sf)
#> Linking to GEOS 3.8.1, GDAL 3.2.1, PROJ 7.2.1
library(terra)
#> terra version 1.4.20

# Fetch the DEM data for Spain
spain_dem = raster::getData("alt", country = "ESP", mask = FALSE)
#> Warning in showSRID(uprojargs, format = "PROJ", multiline = "NO", prefer_proj
#> = prefer_proj): Discarded datum Unknown based on WGS84 ellipsoid in Proj4
#> definition

# Convert the DEM to terra format
spain_dem = rast(spain_dem)
#> Warning: Unhandled datatype=INT2S (GDAL error 1)

# Assign the correct CRS to the DEM: WGS 84
crs(spain_dem) = "EPSG:4326"

# Reduce the resolution by a factor of 20 to speed up calculations
spain_dem = aggregate(spain_dem, fact = 20)

# According to the documentation, terra::distance() will calculate distance
# for all cells that are NA to the nearest cell that are not NA. To calculate
# distance to the coast, we need a raster that has NA values over land and any
# other value over water:
water_mask = is.na(spain_dem)
water_mask[water_mask == 0] = NA

# Use the distance() function on this mask to get distance to the coast
distance_to_coast = distance(water_mask)

# Plot the result
plot(distance_to_coast, main = "Distance to the coast")
```

<sup>Created on 2021-11-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/_04-ex.Rmd
+++ b/_04-ex.Rmd
@@ -112,18 +112,12 @@ cor(two_rasts_df$ndvi, two_rasts_df$ndwi)
 ```
 
 E7. A StackOverflow [post](https://stackoverflow.com/questions/35555709/global-raster-of-geographic-distances) shows how to compute distances to the nearest coastline using `raster::distance()`.
- Retrieve a digital elevation model of Spain, and compute a raster which represents distances to the coast across the country (hint: use `getData()`).
- Note: it may be wise to increase the cell size of the input raster to reduce compute time during this operation.
+Retrieve a digital elevation model of Spain, and compute a raster which represents distances to the coast across the country (hint: use `getData()`).
+Note: it may be wise to increase the cell size of the input raster to reduce compute time during this operation.
 
 ```{r}
 # Fetch the DEM data for Spain
-spain_dem = raster::getData("alt", country = "ESP", mask = FALSE)
-
-# Convert the DEM to terra format
-spain_dem = rast(spain_dem)
-
-# Assign the correct CRS to the DEM: WGS 84
-crs(spain_dem) = "EPSG:4326"
+spain_dem = geodata::elevation_30s(country = "Spain", path = ".", mask = FALSE)
 
 # Reduce the resolution by a factor of 20 to speed up calculations
 spain_dem = aggregate(spain_dem, fact = 20)
--- a/_04-ex.Rmd
+++ b/_04-ex.Rmd
@@ -111,11 +111,33 @@ two_rasts_df = as.data.frame(two_rasts)
 cor(two_rasts_df$ndvi, two_rasts_df$ndwi)
 ```
 
+E7. A StackOverflow [post](https://stackoverflow.com/questions/35555709/global-raster-of-geographic-distances) shows how to compute distances to the nearest coastline using `raster::distance()`.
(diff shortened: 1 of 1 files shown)


Nowosad:
Awesome! 

One thing: could you use `geodata::elevation_30s()` instead of `raster::getData()`?

iod-ine:
@Nowosad Yes! Now it looks a lot cleaner, with no warnings and data conversions.

Attaching a new reprex.

``` r
library(sf)
#> Linking to GEOS 3.8.1, GDAL 3.2.1, PROJ 7.2.1
library(terra)
#> terra version 1.4.20

# Fetch the DEM data for Spain
spain_dem = geodata::elevation_30s(country = "Spain", path = ".", mask = FALSE)

# Reduce the resolution by a factor of 20 to speed up calculations
spain_dem = aggregate(spain_dem, fact = 20)

# According to the documentation, terra::distance() will calculate distance
# for all cells that are NA to the nearest cell that are not NA. To calculate
# distance to the coast, we need a raster that has NA values over land and any 
# other value over water
water_mask = is.na(spain_dem)
water_mask[water_mask == 0] = NA

# Use the distance() function on this mask to get distance to the coast
distance_to_coast = distance(water_mask)

# Plot the result
plot(distance_to_coast, main = "Distance to the coast")
```



<sup>Created on 2021-11-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

iod-ine:
I also removed unnecessary spaces at the beginning of the lines.

Nowosad:
Great!

Nowosad:
@iod-ine Ivan, I also added a two new related exercises along with their solutions - see https://github.com/Robinlovelace/geocompr/commit/e2ee8c73500af54a68a4cce0baf9be3b74d9dbbd. Do you have any suggestion on how I can improve the question in E8?

iod-ine:
@Nowosad I like both the formulation and the solution for E8 :)
E9, however, seems redundant: we are simply restoring the scaled DEM that we added in E8.
I would suggest either ending with E7 and E8 or, if we feel more practice is needed for `terra::distance()`, do something different. Maybe add a use case where we compute distances to a vector, like distance to the capitals for a couple of countries in Europe or something similar?

Nowosad:
Thanks for the suggestions!

Robinlovelace:
Many thanks :pray:

PR #684, opened 2021-11-28
Update data retrieval from spDataLarge

iod-ine:
E4 in C4 uses two rasters that come with `spDataLarge`. To fetch them, the exercise suggests to use the files that come with the package. However, the NDVI file does not exist at the location the description points to.

``` r
ndvi = terra::rast(system.file("raster/ndvi.tif", package = "spDataLarge"))
#> Error: [rast] filename is empty. Provide a valid filename
system.file("raster/ndvi.tif", package = "spDataLarge")
#> [1] ""
```

Here are all the files I have in that folder:

```
dem.tif         landsat.tif     nlcd.tif        nlcd2011.tif    nz_elev.tif     srtm.tif
```

I found that the best way to access both of the files is directly through `spDataLarge` like `spDataLarge::dem` and `spDataLarge::ndvi`. The new version executes without problems:

``` r
library(terra)
#> terra version 1.4.20
dem = rast(spDataLarge::dem)
ndvi = rast(spDataLarge::ndvi)

#1
dem_rcl = matrix(c(-Inf, 300, 0, 300, 500, 1, 500, Inf, 2), ncol = 3, byrow = TRUE)
dem_reclass = classify(dem, dem_rcl)
levels(dem_reclass) = c("low", "medium", "high")
plot(dem_reclass)
```

``` r
#2
zonal(c(dem, ndvi), dem_reclass, fun = "mean")
#>      dem      dem       ndvi
#> 1    low 274.3910 -0.3631755
#> 2 medium 392.0486 -0.2893352
#> 3   high 765.2197 -0.2075496
```

### DIFF
--- a/_04-ex.Rmd
+++ b/_04-ex.Rmd
@@ -43,13 +43,13 @@ nz_height_combined %>%
   na.omit()
 ```
 
-E4. Use `dem = rast(system.file("raster/dem.tif", package = "spDataLarge"))`, and reclassify the elevation in three classes: low (<300), medium and high (>500).
-Secondly, read the NDVI raster (`ndvi = rast(system.file("raster/ndvi.tif", package = "spDataLarge"))`) and compute the mean NDVI and the mean elevation for each altitudinal class.
+E4. Use `dem = rast(spDataLarge::dem)`, and reclassify the elevation in three classes: low (<300), medium and high (>500).
+Secondly, read the NDVI raster (`ndvi = rast(spDataLarge::ndvi)`) and compute the mean NDVI and the mean elevation for each altitudinal class.
 
 ```{r}
 library(terra)
-dem = rast(system.file("raster/dem.tif", package = "spDataLarge"))
-ndvi = rast(system.file("raster/ndvi.tif", package = "spDataLarge"))
+dem = rast(spDataLarge::dem)
+ndvi = rast(spDataLarge::ndvi)
 
 #1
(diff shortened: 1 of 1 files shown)


Nowosad:
@iod-ine I believe you need to update the spDataLarge package to the most recent version - https://github.com/Nowosad/spDataLarge. I have added `ndvi.tif` a few days ago.

iod-ine:
@Nowosad I see, thank you!
The question changes in that case to: what is a better way to retrieve the files? I feel like the updates I suggested are more concise and easier to remember. What do you think? Are there any drawbacks of using this instead of the `system.file`?

Nowosad:
You mean using `spDataLarge::ndvi` vs `terra::rast(system.file("raster/ndvi.tif", package = "spDataLarge"))`?
In the past, I preferred the first solution, but now I prefer the second one. 

Storing spatial data examples as files is just more universal and stable. 

Why? There are several differences between these approaches. The first approach uses an R object -- e.g., `.rda` with raster::RasterLayer. This is a limitation, because: (a) now the data package needs to depend on **raster**, (b) you need to have methods to convert from raster to other classes (if you want to use other classes), (c) if any changes happen to the raster object structure (e.g., like PROJ changes in the last ~2 years), you need to, for example, to rewrite this file. 

Let's say I want to prepare some teaching materials for two groups - one using **stars** and the second **terra**. It is much easier to do it with the second approach.

iod-ine:
@Nowosad Wow, thank you for an awesome explanation. I also prefer the second solution now :)

PR #686, opened 2021-11-30
Fix typos in C7, clarify a sentence in C8

iod-ine:
The changes to chapter 7 are simple typos.

The change in chapter 8 is my attempt to make the sentence more clear. It seems like the sentence started as a comparison of `ggplot2` and `tmap` in terms of variable aesthetics but then ended up just specifying what arguments `tmap` accepts. I feel like splitting these ideas makes sense.


Nowosad:
Thank you, Ivan!

PR #698, opened 2021-12-18
Updates to c5

Robinlovelace:


Robinlovelace:
Heads-up @Nowosad this is good to go. Looking at it, it's a long chapter. I think that splitting out the raster-vector interactions content into a new dedicated chapter could represent a quick win, it's a long chapter and the topic is so important as to merit it's own chapter IMO atm.

Nowosad:
@Robinlovelace Yes -- I think that is a good idea. What do you think @jannes-m ?

Robinlovelace:
Good stuff, I've created a new issue #706 on the topic. Will merge this now to keep things moving.

jannes-m:
Aside from reducing the length of chapter 6, it would be sensible to split out the raster-vector stuff into a new chapter simply because  extracting values from a raster is **not** changing the geometry of an spatial object (and therefore strictly speaking does not belong to c5) and because raster-vector operations are somehow special in any case.

Robinlovelace:
Thanks for the comment Jannes, sounds like there's consensus!

PR #707, opened 2021-12-19
C6 v2 updates rl

Robinlovelace:


Robinlovelace:
This updates C6 mentioning s2 and gets it to MVP status ready for external review (more or less) IMO.

jannes-m (commented):
It's a good idea to mention s2 early on. Still, we should keep in mind that sf has already supported  computations on an ellipsoid with the help of lwgeom before. The main difference now is that it uses s2, i.e., computations on the sphere, by default, when given an unprojected spatial object.

Robinlovelace:
This PR has been replaced by #709. Leaving this PR open until that one gets merged, we could merge the two PRs but probably simplest to merge #709 and delete this branch.

PR #710, opened 2021-12-22
Update 01-introduction.Rmd

e-clin:
Corrected a minor spelling mistake - in computer science, it should be "disk" instead of "disc", right?


Robinlovelace (approved):
Never thought about it until now but it seems you're right: https://www.merriam-webster.com/words-at-play/disc-vs-disk-usage-history-spelling 

Many thanks!

e-clin:
> Never thought about it until now but it seems you're right: https://www.merriam-webster.com/words-at-play/disc-vs-disk-usage-history-spelling
> 
> Many thanks!

Cheers! I didn't pick it up at the first glance, either.

Just started reading the book - thanks very much for sharing your hard work! 👍

Issue #713, opened 2021-12-30
Update Section 7.4


Robinlovelace:
https://geocompr.robinlovelace.net/reproj-geo-data.html#reproj-vec-geom

PR #725, opened 2022-01-22
CRS updates

Robinlovelace:
See #713

### DIFF
--- a/07-reproj.Rmd
+++ b/07-reproj.Rmd
@@ -47,12 +47,12 @@ But which one is correct?
 
 The short answer is the third definition is correct: `EPSG:4326` is understood by `sf` (and by extension `stars`) and `terra` packages covered in this book, has the benefit of being human readable (search for it online and you will find a web page describing the CRS titled EPSG:4326 at on the website [epsg.io](https://epsg.io/4326), for example), and is future proof.
 The more concise identifier `4326` is understood by `sf`.
-However, **we urge users to use the more explicit `AUTHORITY:code` representation to prevent ambiguity and to provide context**.
+However, **we urge users to use the more explicit `AUTHORITY:CODE` representation to prevent ambiguity and to provide context**.
 
 A longer answer is that more detail is needed.
 None of three descriptions above contain sufficient information about WGS84 or any CRS for unambiguous transformations: due to the complexity of CRSs, it is not possible to capture all relevant information about them in such short text strings.
-For this reason, the Open Geospatial Consortium (OGC, which also developed the simple features specification that the `sf` package implements) developed an open standard format for describing CRSs that is called WKT (Well Known Text) in a [100+ page document](https://portal.opengeospatial.org/files/18-010r7) that "defines the structure and content of a text string implementation of the abstract model for coordinate reference systems described in ISO 19111:2019" [@opengeospatialconsortium_wellknown_2018].
-The WKT representation of the WGS84 CRS is as follows:
+For this reason, the Open Geospatial Consortium (OGC, which also developed the simple features specification that the `sf` package implements) developed an open standard format for describing CRSs that is called WKT (Well Known Text) in a [100+ page document](https://portal.opengeospatial.org/files/18-010r7) that "defines the structure and content of a text string implementation of the abstract model for coordinate reference systems described in ISO 19111:2019" [@opengeospatialconsortium_wellknown_2019].
+The WKT representation of the WGS84 CRS, which has the **identifier** `EPSG:4326` is as follows:
 
 <!-- Source: https://spatialreference.org/ref/epsg/4326/prettywkt/ -->
 <!-- ``` -->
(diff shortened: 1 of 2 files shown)


Robinlovelace:
Heads-up @Nowosad the intention with this PR is to complete Chapter and therefore (I think) most of the stuff needed to submit Part 1 for review. Sorry it's taken so long. Key points: we should not mention 'WKT2' as it's confusing, WKT2 is and in the future will be synonymous with WKT representation of CRSs. I have read and linked to key related docs on CRSs. One thing that surprised me: it does not seem that the AUTHORITY:CODE string convention in `EPSG:4326` for example is universally accepted. The only standard is WKT, there are still different ways of identifying WKT strings as outlined below and now in this PR.

https://docs.opengeospatial.org/is/18-010r7/18-010r7.html#37 now given the proper citation it deserves, the most important doc here

The QGIS docs which show 5 CRS identifier types: https://docs.qgis.org/3.16/en/docs/pyqgis_developer_cookbook/crs.html?highlight=srid

PostGIS docs which is where I think the term 'SRID' comes from. I think 'identifier' is better for our purposes: https://postgis.net/workshops/postgis-intro/projection.html

As others have said this is a minefield. But I'm confident the updated text is accurate and better. Needed to rework these earlier sections before getting to 7.4. Any comments v. welcome.

Nowosad:
Hi @Robinlovelace,
thanks for opening the discussion:

1. WKT2 -- as I mentioned in the text -- this is just a dialect of WKT. If you just want to use WKT consistently in the text (instead of WKT2) that is fine by me. However, we probably should mention this dialect once (e.g., in chapter 2) and say that we will use just WKT from now on...
2. "It does not seem that the AUTHORITY:CODE string convention in EPSG:4326 for example is universally accepted." -- I think PROJ and GDAL universally accept it -- and that is good for us. You just need to remember that only WKT is used for storing CRSs, and the AUTHORITY:CODE convention is mostly for setting, changing, and transforming CRSs.
3. The QGIS code, I think, is just like sf or terra -- these examples just use PROJ API plus some specific QGIS code. See our Section 7.3:

image

4. Correct me if I am wrong, by SRID means Spatial Reference Identifier. I am not very attached to this name, so we could use just "identifier" instead. The question, however, is if that is enough not to confuse the readers (would they understand that by identifier, we mean CRS identifies/SRID?)
5. Let me know when the text is ready for my review.

J

Robinlovelace:
Hi @Nowosad all good questions and it's a minefield so worth discussing. Responses soon...

Robinlovelace:
Regarding your point 1:

> WKT2 -- as I mentioned in the text -- this is just a dialect of WKT. If you just want to use WKT consistently in the text (instead of WKT2) that is fine by me. However, we probably should mention this dialect once (e.g., in chapter 2) and say that we will use just WKT from now on...

The fact that the WKT representation has changed over time is not relevant for most users and will be come less relevant, tending towards obsolete, in the coming years. Good idea to mention it. I think a footnote in C7 would be the best place for that, e.g. here: 

https://github.com/Robinlovelace/geocompr/blob/fa05ae039b2e7f45dc9ef0059d5b0adb95a52b7d/07-reproj.Rmd#L100-L107

That's my default plan, it fits well there because that footnote also mentions the increasingly obsolete `proj4` string representations. Note to self: need to find and change all references to WKT2 for consistency (hence putting in as a PR to ensure it's not breaking the logic of other parts of the book).

Robinlovelace:
> I think PROJ and GDAL universally accept it -- and that is good for us. You just need to remember that only WKT is used for storing CRSs, and the AUTHORITY:CODE convention is mostly for setting, changing, and transforming CRSs.

Good idea to focus on what PROJ and GDAL accept, I was looking (perhaps too much) at the PostGIS docs. Yes of course the identifier is only a look-up and that the CRS is stored in the WKT string. That is now more clear I think. If we had links showing that PROJ/GDAL accept the `AUTHORITY:CODE` representation (which I called `AUTHORITY:code`, plan to change as all upper looks clearer) that would help, do you have any links to docs in PROJ where identifiers are used?

Robinlovelace:
> The QGIS code, I think, is just like sf or terra -- these examples just use PROJ API plus some specific QGIS code. See our Section 7.3:

Good point on that, I think clarifying that and adding a cross reference to the subsequent section is a good idea. However, I think the QGIS example shows that there are alternative CRS identifiers other than the 'SRID'.

Robinlovelace:
> Correct me if I am wrong, by SRID means Spatial Reference Identifier. I am not very attached to this name, so we could use just "identifier" instead. The question, however, is if that is enough not to confuse the readers (would they understand that by identifier, we mean CRS identifies/SRID?)

Yes according to PostGIS documentation. But where else is that term used? I'm not sure that in PostGIS docs they would say that `EPSG:4326` is an SRID. Do you have more info on that (I tried to find other places where SRID was used but could not find much)? 

I think if we define CRS identifiers early and clearly, with reference to the OGC standard which calls them identifiers, it will be clear to readers. Pros and cons. ATM I'm thinking the advantages of just 'identifier' used by OGC outweigh the advantages of `SRID`, used in PostGIS docs (and possibly elsewhere?).

Robinlovelace:
Another observation: there is really good documentation here which mentions identifiers but avoids the acronym SRID: https://proj.org/operations/operations_computation.html However there is mention of SRIDs here: https://proj.org/faq.html

Robinlovelace:
But that seems to be the only mention of the term: https://www.google.com/search?q=srid+%22spatial+reference%22+site%3Aproj.org

Robinlovelace:
In GDAL docs there seems to be no mention or srids: https://www.google.com/search?q=srid+%22spatial+reference%22+site%3Agdal

Nowosad (approved):

PR #746, opened 2022-02-06
improves s2example

Nowosad:


Nowosad:
Maybe we should also mention maxcells somewhere, such as in `india_buffer_with_s2 = st_buffer(india, 1, max_cells = 10000)`.

Robinlovelace:
> Maybe we should also mention maxcells somewhere, such as in `india_buffer_with_s2 = st_buffer(india, 1, max_cells = 10000)`.

I think that would be too much duplication of effort, as long as people are signposted to the projections chapter they will find out there.

Robinlovelace (approved):
:+1:

PR #749, opened 2022-02-10
More badges, fewer remotes

Robinlovelace:


Robinlovelace:
This commit adds new social GitHub info that we should be proud of:

image

And, the bit I'm less sure about, fewer installation instructions because it builds on CRAN versions of everything, and {geocompkg} installs {spDataLarge} now I think: 

image

Nowosad:
👍🏻 
The terra and spData dev installations can be removed for sure. I am less sure about spDataLarge and tmap... However, we are using a docker with geocompkg only and it works fine. So I think we can merge this PR.

Robinlovelace:
We'll find out soon if people open issues. I think it will be fine, but let's see!

PR #750, opened 2022-02-13
delete l119 as it is already said in l110

defuneste:
I do not think we should keep  L119 because of L100 already said it.


Robinlovelace (approved):
Good deduplication work. Many thanks!

PR #758, opened 2022-02-24
sf_use_s2("FALSE") -> sf_use_s2(FALSE)

kwhkim:


Robinlovelace (approved):
Definite fix, not sure how that got in there (probably my fault!) and very grateful for this contribution. Thank you @kwhkim :pray:

Issue #766, opened 2022-03-06
Some ideas for exercices on ch04 on spatial predicates


defuneste:
Spatial predicates are hard and I think adding them in exercices can help. This is some "raw" ideas that can be explored. I need to find a data set to display  "st_crosses".

``` r
## Load package and pick some data 
library(sf)
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1
library(spData)

# One US state
Colorado = us_states[us_states$NAME == "Colorado",]

plot(us_states$geometry)
plot(Colorado$geometry, col = 2, add = TRUE)
```

![](https://i.imgur.com/xghbbCx.png)

``` r
# What are the neighbouring states of Colorado? 
sgbd1 = st_touches(
    Colorado, 
    us_states)
us_states$NAME[unlist(sgbd1)]
#> [1] "Arizona"    "Kansas"     "Oklahoma"   "Nebraska"   "New Mexico"
#> [6] "Utah"       "Wyoming"

# how many states do we get and why? 
us_states$NAME[unlist(st_intersects( Colorado, 
                                us_states))]
#> [1] "Arizona"    "Colorado"   "Kansas"     "Oklahoma"   "Nebraska"  
#> [6] "New Mexico" "Utah"       "Wyoming"

# to be sure !
us_states$NAME[unlist(st_equals( Colorado, 
                                us_states))]
#> [1] "Colorado"

# Try a pattern:
small_neighbour = "****0****"

us_states$NAME[unlist(st_relate(Colorado, 
                           us_states,
                           pattern = small_neighbour))]
#> although coordinates are longitude/latitude, st_relate_pattern assumes that they are planar
#> [1] "Arizona"

# in the pattern change 0 for 1, what happens? why?   
small_neighbour = "****1****"

us_states$NAME[unlist(st_relate(Colorado, 
                           us_states,
                           pattern = small_neighbour))]
#> although coordinates are longitude/latitude, st_relate_pattern assumes that they are planar
#> [1] "Colorado"   "Kansas"     "Oklahoma"   "Nebraska"   "New Mexico"
#> [6] "Utah"       "Wyoming"

# Bonus why do you have a warning with st_relate and not with the other spatiale predicates ? 
```

<sup>Created on 2022-03-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Robinlovelace:
[...]

PR #770, opened 2022-03-18
Update answer to c3 to include tidyverse solution

Robinlovelace:


Robinlovelace:
And your review/input greatly appreciated also @defuneste. Many thanks for the idea, this gets us about half way to implementing it! I think a straight line between states could work well as a next step but happy with anything.

Robinlovelace:
Merging for now to keep it moving forward, unfinished business in #766, can pick up on that.

PR #782, opened 2022-04-21
Review of ch12

Nowosad:
Hi @jannes-m -- I read the whole chapter (before the most recent changes) and learned a lot! 

I also added a number of rather small changes and have a few comments:
1. I would suggest cleaning most of the commented lines in the chapter. 
2. Line 145 -- contains a link to datacamp materials. Given the problematic history of the company, I suggest to replace it.
3. Line 463 -- You say "...as expected...". I think it would not be expected by many people not familiar with this topic. I would suggest adding a sentence or a few explaining why this is expected.
4. Line 631 -- I would suggest adding an example runtime (e.g., "on a modern laptop it took XX seconds")

PR #783, opened 2022-04-24
Update with consistent bullet points/colons

Robinlovelace:
As discussed here: https://github.com/Robinlovelace/geocompr/pull/782


Robinlovelace:
Heads-up Jakub I think this is now more consistent. Any comments, or other examples that could be make more standard/consistent let me know.

Nowosad (approved):
perfecto

PR #792, opened 2022-05-07
Add links to GDAL, PROJ, and GEOS

andtheWings:
...to their Ch. 2 descriptions


Robinlovelace (approved):
Good to have links. Attitude before was to use them sparingly due to typesetting considerations. Those considerations no longer hold so big :+1: from me on this, many thanks!

PR #793, opened 2022-05-07
Rev15

Nowosad:


jannes-m (commented):
Thanks for improving the prose and for converting my lattice code to tmap code, very much appreciated @Nowosad. I will only use austral winter again instead of saying only winter. This reminds the reader hopefully, that the study area is located on the southern hemisphere.

Nowosad:
👍🏻

Issue #798, opened 2022-05-13
clarrification on ch05 text describing table 5.1


defuneste:
Hi @Robinlovelace and @Nowosad 

I have trouble understanding few sentences describing table 5.1 in chp 05 . 

This one is difficult: https://github.com/Robinlovelace/geocompr/blob/main/05-geometry-operations.Rmd#L524

> Each input simple feature object with only one element (first column) is transformed directly into another geometry type.  

And this one is maybe incomplete : 

>  You can see that, for example, when you cast a multipoint consisting of five pairs of coordinates into a point.

here : https://github.com/Robinlovelace/geocompr/blob/main/05-geometry-operations.Rmd#L524

++

Robinlovelace:
As discussed on Discord, a good next step could be to chat it over. I would be up for that at some point next week. 

Link to the content online: https://geocompr.robinlovelace.net/geometric-operations.html#tab:sfs-st-cast

And permalink to the source code:

https://github.com/Robinlovelace/geocompr/blob/fe65a3a0255b06bbb4860fa91009496f5d124fbb/05-geometry-operations.Rmd#L519-L549

Nowosad:
Hi guys -- please look at https://github.com/Robinlovelace/geocompr/pull/799 and let me know what you think.

Robinlovelace:
Looking better. Suggested changes building on your changes here Jakub: https://github.com/Robinlovelace/geocompr/pull/800/files

A couple of points: in the table it is not clear what each input and output is, are they `sfc` objects of different lengths (number of rows when in `sf` data frames)?

And the word 'element' is not clear. I think '`sfg` object' could be clearer.

Nowosad:
The inputs are based on the objects from https://github.com/geocompr/geocompr.github.io/blob/source/content/post/wip/casting-geometries-extended.Rmd.
So -- the first column represents input objects (with various geometry types), and all of the rest columns is "what will happen if you type transform the input object to some other geometry".

Robinlovelace:
According to comments by @defuneste here https://github.com/Robinlovelace/geocompr/pull/799#issuecomment-1126802887 this seems fixed. Thanks for reporting and for the quick fix!

PR #799, opened 2022-05-14
related to #798

Nowosad:


05-geometry-operations.Rmd
```diff
@@ -516,9 +516,9 @@ t = cast_all(multipoint)
...
 ```
 
-Geometry casting of simple features geometry column (`sfc`) and simple features objects works the same as for single geometries in most of the cases. 
+Geometry casting of simple features geometry column (`sfc`) and simple features objects works the same as for `sfg` in most of the cases. 
 One important difference is the conversion between multi-types to non-multi-types.
-As a result of this process, multi-objects are split into many non-multi-objects.
+As a result of this process, multi-objects of `sfc` or `sf` are split into many non-multi-objects.
 
 Table \@ref(tab:sfs-st-cast) shows possible geometry type transformations on simple feature objects.
 Each input simple feature object with only one element (first column) is transformed directly into another geometry type.
```
Robinlovelace: I think this could be explained more clearly.

defuneste (commented):
This is way better, specially L507. Let me try to improve a bit L524 (and see now, if I understand it correctly).  

> Each input simple feature object with only one element (first column) is transformed directly into another geometry type.

---> 

When a simple feature object with just one element is used in input it is transformed directly into another geometry type.


And L527

> You can see that, for example, when you cast a multipoint consisting of five pairs of coordinates into a point.

---> 

You can see that, for example, when you cast a multipoint consisting of five pairs of coordinates into a point (one pair of coordinates). 

Correct ?

Robinlovelace:
To keep up momentum I'm going to merge this. We can see how it looks in the updated text and then decide if it fixes the issue. Any further comments welcome @defuneste. I think the table still could be explained better in terms of exactly what the inputs in the first column are. Any ideas on that? Far out idea is to create a small plot representing each input...

Robinlovelace (approved):

defuneste:
https://github.com/Robinlovelace/geocompr/blob/main/05-geometry-operations.Rmd#L524-L527

This seems good to me and showed that I misunderstood the first version!  

I do not see how we can improve the table. An exercise on it could be an option but E6 is already a bit on it.

PR #801, opened 2022-05-29
Fix typo

cuixueqin:


Robinlovelace (approved):
Definite fix, well spotted, many thanks!

PR #804, opened 2022-06-05
WTK to WKT

adambhouston:
just corrected spelling in line 81 - WTK was written instead of WKT.

first-time github user, please pardon any etiquette breaches!


Nowosad:
@adambhouston Thank you a lot, Adam! Good catch.

PR #805, opened 2022-06-05
Update 12-spatial-cv.Rmd

babayoshihiko:
When I ran the code, `rr_glm` did not exist. Maybe `rr_spcv_glm`?

The code chunk has eval=F, so it will not affect building.


jannes-m (approved):
Thanks for spotting

Robinlovelace (approved):
A danger of eval=FALSE. Definite fix :+1:

PR #806, opened 2022-06-08
Fix typo

cuixueqin:


Robinlovelace (approved):
That is correct, not only was the existing text a typo. It was incorrect! I approve this change. Please take a look @Nowosad also, look good to you, should we make additional changes?

Nowosad (approved):

PR #807, opened 2022-06-08
Replace top_n() with slice_max()

cuixueqin:
Replace top_n() with slice_max() since it has been superseded in favour of slice_min()/slice_max().


Robinlovelace (approved):
I hadn't seen this but, looking at the [documentation](https://dplyr.tidyverse.org/reference/top_n.html), see you're right @cuixueqin. Many thanks for another improvement to these open teaching materials.

PR #822, opened 2022-06-30
rasters?

babayoshihiko:
Is this just a typo?


Robinlovelace (approved):
Yes defo typo fix, hard to spot hiding behind the 's but it's there for sure. Thanks!

PR #829, opened 2022-07-06
footnote starting line 114

babayoshihiko:
The footnote starting at line 114 ends at line 124, containing two paragraphs.

This is currently the footnote 27 in https://geocompr.robinlovelace.net/reproj-geo-data.html#crs-in-r . However the second paragraph does not  appear in the foot note.

Reading the footnote, the footnote is supposed to be just the first part. So, `]` should be in line 117?


Nowosad (approved):

Nowosad:
Thank you @babayoshihiko !

PR #831, opened 2022-07-07
Figure 9.1

babayoshihiko:
This chapter is now Chapter 9, so changing the Figure number from 8.1 to 9.1.


Nowosad:
Thank you @babayoshihiko -- we still plan to rewrite the whole chapter 9, but this is a good fix for the moment.

babayoshihiko:
I've started reading/translating Chapters 9/10. Maybe I should stay Chapters 1-8?

PR #843, opened 2022-08-08
Missing opening ^[ in 10-gis.Rmd

babayoshihiko:
The closing  bracket in line 93 had no corresponding opening bracket.


Robinlovelace (approved):
Good point, the original was commented out:

```
<!-- Though not covered here, it is worth being aware of the interface to ArcGIS\index{ArcGIS}, a proprietary and very popular GIS software, via **RPyGeo**.^[By the way, it is also possible to use R from within Desktop GIS software packages.  -->
```

PR #868, opened 2022-09-06
Tweak s2geo figure, close #867

### LINKED ISSUE #867: Updates related to diagram in Chapter 7
Related to this bit of code:

https://github.com/Robinlovelace/geocompr/blob/60d832e58ff304f1ac21393762390ad25d4115da/07-reproj.Rmd#L259

- [x] ~Add DiagrammeR Suggests in **geocompkg**~
- [x] Ensure text fits in boxes (maybe convert to png)
- [ ] Mention that `sf_use_s2()` setting will mean GEOS is used on lon/lat data
- [ ] Emphasise that the results are different when S2 is used? I think having 2 outputs side-by-side could help highlight that, currently the single result could lead people to believe that the result will be the same regardless of which geometry engine is used but in fact results can vary dramatically and this is an important point

Re. package dependency it caused `bookdown::serve_book()` to fail on my laptop. Although strangely GitHub Actions are working fine.

Robinlovelace:


Nowosad:
One last thing -- can you add the image in a slightly higher resolution?

Robinlovelace:
> One last thing -- can you add the image in a slightly higher resolution?

I tried as you can see in the commits above, but failed.

Nowosad (approved):

PR #888, opened 2022-10-23
Update 13-transport.Rmd

Robinlovelace:
Update section on route/leg/segment level outputs   

Heads-up @temospena, seem reasonable? See https://github.com/ipeaGIT/r5r/issues/298


temospena:
It's good to have the 3 distinction.
I think the second should be "leg", as a leg journey. Maybe confusing to call bother segmentos (2 & 3)

Robinlovelace:
> I think the second should be "leg", as a leg journey. Maybe confusing to call bother segmentos (2 & 3)

Oh yes, just noticed that typo! Will fix...

PR #899, opened 2022-12-23
Loc jn reb

Nowosad:
@jannes-m -- my small changes to the chapter are attached.


14-location.Rmd
```diff
@@ -162,10 +158,10 @@ Be careful with geographic CRS\index{CRS!geographic} where grid cell areas const
 ```
 
 ```{r census-stack, echo=FALSE, fig.cap="Gridded German census data of 2011 (see Table \\@ref(tab:census-desc) for a description of the classes).", fig.scap="Gridded German census data."}
-knitr::include_graphics("figures/08_census_stack.png")
+knitr::include_graphics("figures/14_census_stack.png")
 ```
 
-The next stage is to reclassify the values of the rasters stored in `input_ras` in accordance with the survey mentioned in Section \@ref(case-study), using the **raster** function `reclassify()`, which was introduced in Section \@ref(local-operations)\index{map algebra!local operations}.
+The next stage is to reclassify the values of the rasters stored in `input_ras` in accordance with the survey mentioned in Section \@ref(case-study), using the **terra** function `classify()`, which was introduced in Section \@ref(local-operations)\index{map algebra!local operations}.
```
jannes-m, line 164: good catch

14-location.Rmd
```diff
@@ -144,8 +140,8 @@ knitr::kable(tab,
 ## Create census rasters
  
 After the preprocessing, the data can be converted into a `SpatRaster` object (see Sections \@ref(raster-classes) and \@ref(raster-subsetting)) with the help of the `rast()` function.
-Setting its `type` argument to `xyz`, it requires an input data frame whose columns `x` and `y` represent coordinates on a regular grid.
-All the remaining columns (here: `pop`, `women`, `mean_age`, `hh_size`) will serve as input for the raster layers (Figure \@ref(fig:census-stack); see also `code/14-location-figures.R` in our github repository).
+When setting its `type` argument to `xyz`, an input data frame whose columns `x` and `y` will represent coordinates on a regular grid.
```
jannes-m, line 143: Sentence is somehow wrong, I'll rewrite this.

14-location.Rmd
```diff
@@ -215,7 +211,7 @@ reclass
 
 ## Define metropolitan areas
 
-We define metropolitan areas as pixels of 20 km^2^ inhabited by more than 500,000 people.
+We decided to define metropolitan areas as pixels of 20 km^2^ inhabited by more than 500,000 people.
```
jannes-m, line 214: Does decide add something valuable to the meaning of the sentence -:)? I would prefer the first version but can live with this as well.

14-location.Rmd
```diff
@@ -234,32 +230,21 @@ Plotting this reveals eight metropolitan regions (Figure \@ref(fig:metro-areas))
...
 
 ```{r 14-location-13, warning=FALSE, message=FALSE}
-polys = pop_agg |> 
+metros = pop_agg |> 
   terra::patches(directions = 8) |>
   terra::as.polygons() |>
   sf::st_as_sf()
 ```
 
-`polys` now features a column named `patches` which indicates to which metropolitan region each polygon belongs and which we will use to dissolve\index{dissolve} the polygons into coherent single regions (see also Section \@ref(geometry-unions)):
```
jannes-m, line 246: Why have you deleted this part? We need the dissolved regions for the reverse geocoding later on.

14-location.Rmd
```diff
@@ -321,7 +307,7 @@ shops = purrr::map(metro_names, function(x) {
   points = osmdata::osmdata_sf(query)
   # request the same data again if nothing has been downloaded
   iter = 2
-  while (nrow(points$osm_points) == 0 & iter > 0) {
+  while (nrow(points$osm_points) == 0 && iter > 0) {
```
jannes-m, line 310: :+1:

14-location.Rmd
```diff
@@ -343,7 +329,7 @@ if (any(ind)) {
 }
 ```
 
-To make sure that each list element (an `sf`\index{sf} data frame) comes with the same columns^[This is not a given since OSM contributors are not equally meticulous when collecting data.] we only keep the `osm_id` and the `shop` columns with the help of the `map_dfr` loop which additionally rowbinds all shops into one large `sf`\index{sf} object.
+To make sure that each list element (an `sf`\index{sf} data frame) comes with the same columns^[This is not a given since OSM contributors are not equally meticulous when collecting data.] we only keep the `osm_id` and the `shop` columns with the help of the `map_dfr` loop which additionally combines all shops into one large `sf`\index{sf} object.
```
jannes-m, line 332: yes, you are right, rowbinds is not a word, thanks!

14-location.Rmd
```diff
@@ -366,8 +352,7 @@ The result of the subsequent code chunk is therefore an estimate of shop density
 ```{r 14-location-25, message=FALSE, warning=FALSE}
 shops = sf::st_transform(shops, st_crs(reclass))
 # create poi raster
-poi = terra::rasterize(x = terra::vect(shops),
-                       y = reclass, field = "osm_id", fun = "length")
+poi = terra::rasterize(x = shops, y = reclass, field = "osm_id", fun = "length")
```
jannes-m, line 355: are you sure we don't need to convert an sf-object into an vect-object? Ok, let's give it a try and see if it works -:)

code/14-location-figures.R
```diff
@@ -57,38 +54,19 @@ input_tidy = dplyr::mutate(
 input_ras = terra::rast(input_tidy, type = "xyz", crs = "EPSG:3035")
 
 # reproject German outline
-ger = terra::project(ger, crs(input_ras))
+ger = st_as_sf(terra::project(ger, crs(input_ras)))
 
 # 2.2 Create figure========================================
 #**********************************************************
+tm_1 = tm_shape(input_ras) +
```
jannes-m, line 61: And this does the same as the `spplot` code?? This is impressive, and I should definitely look more into **tmap**.

code/14-location-figures.R
```diff
@@ -114,94 +92,27 @@ pop_agg = terra::aggregate(reclass$pop, fact = 20, fun = sum, na.rm = TRUE)
 # just keep raster cells with more than 500,000 inhabitants
 polys = pop_agg[pop_agg > 500000, drop = FALSE] 
 # convert all cells belonging to one region ino polygons
-polys = polys |>
+metros = polys |>
   terra::patches(directions = 8) |>
   terra::as.polygons() |>
   st_as_sf()
-# dissolve
-metros = polys |>
```
jannes-m, line 122: same comment as above, I think we need to dissolve

code/14-location-figures.R
```diff
@@ -114,94 +92,27 @@ pop_agg = terra::aggregate(reclass$pop, fact = 20, fun = sum, na.rm = TRUE)
...
-# save the output
-ggplot2::ggsave(filename = "figures/08_metro_areas.png",
-                plot = arrangeGrob(p_3, ncol = 1),
-                width = 3, height = 4)
+metros$names = c("Hamburg", "Berlin", "Düsseldorf", "Leipzig",
+                 "Frankfurt am Main", "Nürnberg", "Stuttgart", "München")
+
+metros_points = st_centroid(metros)
+
+tm_2 = tm_shape(pop_agg/1000) +
```
jannes-m, line 104: and again, the same comment as above, the **tmap** code looks much more concise and better to follow! Thanks a lot! I need to have  a look at the new figures!

jannes-m (commented):
@Nowosad , thanks for your review which I very much appreciate especially the rewrite of the figures using **tmap**. Let's discuss the dissolving issue, maybe I am wrong and it is unnecessary?

14-location.Rmd
jannes-m: excellent, you are right, it works :+1:

code/14-location-figures.R
jannes-m: would it be possible to print the metro names in italics and with a smaller font?

14-location.Rmd
jannes-m: And again, you are right, the dissolving operation is no longer necessary. Thanks for spotting!

code/14-location-figures.R
jannes-m: apparently, we don't. Many thanks again for spotting, great catch!!!

14-location.Rmd
Nowosad: 👍🏻

14-location.Rmd
Nowosad: Maybe it is only me, but I would interpret these two sentences differently:

"We define metropolitan areas as pixels of 20 km^2^ inhabited by more than 500,000 people." (as we ALWAYS define it this way)

"We decided to define metropolitan areas as pixels of 20 km^2^ inhabited by more than 500,000 people." (as we defined it this way for the purpose of the study)

14-location.Rmd
Nowosad: Yep -- `as.polygons()` dissolves borders by default.

14-location.Rmd
Nowosad: 👍🏻

code/14-location-figures.R
Nowosad: I tried to made new figures as similar to the old ones as possible (the legend position is the biggest difference).

code/14-location-figures.R
Nowosad: I made the font smaller and, at the same time, discovered a bug in {tmap} preventing me from printing the names in italics (https://github.com/r-tmap/tmap/issues/686).

14-location.Rmd
jannes-m: Ok, good point, I haven't looked at it this way. So I will rewrite to: "We deliberately define..."

14-location.Rmd
jannes-m: Excellent, I haven't been aware of this, and I think it was different with the raster command back then. Anyways, thanks a lot for noting, makes the code more concise!

code/14-location-figures.R
jannes-m: excellent work, Jakub, thanks a lot!

code/14-location-figures.R
jannes-m: :+1:

jannes-m:
Jakub, thanks a ton for the great review and for improving the chapter in terms of prose and for getting it up to speed with more modern approaches to plotting (**tmap** instead of `spplot()`).

jannes-m:
And merry Christmas to you and @Robinlovelace !! We encountered a little watery surprise in our cellar today. The water had a height of 10-15cm, great timing. So I cleaned out the cellar, found a lot of stuff that should go to the garbage in any case and it gives us a reason to scrub it properly :smile: I hope you didn't encounter such surprises and enjoyed Christmas with your little ones and your families!

Nowosad:
Thanks @jannes-m. I hope the "cellar issue" is now resolved.

PR #902, opened 2022-12-25
Removing remaining resolution argument

darrellcarvalho:
Removing the resolution argument from the code block defining the categorical raster, to keep it consistent with decision in the code block defining elev raster.


jannes-m:
Thanks a lot!

Issue #895, opened 2022-12-16
Unclear meaning


jannes-m:
https://github.com/Robinlovelace/geocompr/blob/14acbbec6ace79f2c8dcc7d0efe7b3f137fb2192/04-spatial-operations.Rmd#L679

Yes, this works for projected and unprojected data. In the latter case, however, a value of 20 would refer to a distance of 20 degrees, not 20 m. Shouldn't we point out that the distance value has to be given in map units, i.e., in the unit as used by the used crs?

Robinlovelace:
> a value of 20 would refer to a distance of 20 degrees, not 20 m

I think not after **sf** v1.0.0:

``` r
# Aim: test st_within_distance
library(sf)
#> Linking to GEOS 3.10.2, GDAL 3.4.1, PROJ 8.2.1; sf_use_s2() is TRUE
central_london = spData::lnd[33, ]
buffer_5km = sf::st_buffer(central_london, dist = 5000)
# st_intersects
intersects = spData::lnd[central_london, ]
plot(sf::st_geometry(intersects))
```

![](https://i.imgur.com/fVXbqEC.png)

``` r
within_5km = spData::lnd[central_london, , op = sf::st_is_within_distance, dist = 5000]
plot(sf::st_geometry(within_5km))
plot(buffer_5km, lwd = 4, col = NA, add = TRUE)
#> Warning in plot.sf(buffer_5km, lwd = 4, col = NA, add = TRUE): ignoring all but
#> the first attribute
```

![](https://i.imgur.com/Wy2QMj7.png)

<sup>Created on 2022-12-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Worth clarifying that this statement assumes objects have the correct CRS?

jannes-m:
Interesting. I will have another look at it on Monday!

jannes-m:
Hey Robin,
you were right, by default, the `dist` argument seems to expect meters (though I couldn't find this in the documentation). Hence, it might be a good idea to be specific by explicitly stating what kind of unit one is using, e.g., `units::set_units(5000, "m")`. Interestingly, the units of the `dist` argument are again based on the crs when turning of s2.

``` r
library(sf)
#> Linking to GEOS 3.9.1, GDAL 3.4.3, PROJ 7.2.1; sf_use_s2() is TRUE
central_london = spData::lnd[33, ]
# st_crs(central_london)  # 4326
[...]

PR #908, opened 2023-01-01
avoid potential ambiguity

jannes-m:
see #895

### DIFF
--- a/04-spatial-operations.Rmd
+++ b/04-spatial-operations.Rmd
@@ -677,8 +677,7 @@ if (knitr::is_latex_output()){
 Imagine that we need to join the `capacity` variable in `cycle_hire_osm` onto the official 'target' data contained in `cycle_hire`.
 This is when a non-overlapping join is needed.
 The simplest method is to use the binary predicate `st_is_within_distance()`, as demonstrated below using a threshold distance of 20 m.
-Note that this works with projected data and also with unprojected data.
-In the latter case, however, one needs to make sure that the spherical geometry engine (s2) is switched on which it is by default in **sf** (see section \@ref(s2)).
+One can set the threshold distance in metric units also for unprojected data (e.g. lon/lat CRSs such as WGS84), if the spherical geometry engine (s2) is enabled, as it is in **sf** by default (see section \@ref(s2)).
 
 ```{r 04-spatial-operations-24}
 sel = st_is_within_distance(cycle_hire, cycle_hire_osm, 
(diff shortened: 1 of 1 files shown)


04-spatial-operations.Rmd
```diff
@@ -676,10 +676,13 @@ if (knitr::is_latex_output()){
 
 Imagine that we need to join the `capacity` variable in `cycle_hire_osm` onto the official 'target' data contained in `cycle_hire`.
 This is when a non-overlapping join is needed.
-The simplest method is to use the topological operator `st_is_within_distance()`, as demonstrated below using a threshold distance of 20 m (note that this works with projected and unprojected data).
+The simplest method is to use the topological operator `st_is_within_distance()`, as demonstrated below using a threshold distance of 20 m.
```
Robinlovelace, line 679: Another thought: is this more of a binary predicate than a topological operator? "Topological operator" implies to me that it's operating on, change, the topology of the geometry, whatever that means...

Robinlovelace: Of course that comment applies equally to the old text as to your new text, just a rather late-in-the-day thought!

04-spatial-operations.Rmd
```diff
@@ -676,10 +676,13 @@ if (knitr::is_latex_output()){
...
+Note that this works with projected data and also with unprojected data.
+In the latter case, however, one needs to make sure that the spherical geometry engine (s2) is switched on which it is by default in **sf** (see section \@ref(s2)).
```
Robinlovelace, line 681: :+1:

Robinlovelace: Shorter version could be:

```
For unprojected (e.g. lon/lat CRSs such as WGS84) the spherical geometry engine (s2) must be enabled for `st_within_distance()` to work,, as it is in **sf** (see section \@ref(s2)).
```

Robinlovelace (approved):

04-spatial-operations.Rmd
jannes-m: > Another thought: is this more of a binary predicate than a topological operator? "Topological operator" implies to me that it's operating on, change, the topology of the geometry, whatever that means...

Yes, let's use binary, see https://r-spatial.org/book/03-Geometries.html#sec-de9im

04-spatial-operations.Rmd
jannes-m: Yes, way better, let's use your version. I will update as soon as I am back on my laptop, for now, I need to get some sleep. Happy new year by the way 😊

Robinlovelace: Happy New Year :fireworks:

jannes-m: I slightly rewrote: 

```
One can set the threshold distance in metric units also for unprojected data (e.g. lon/lat CRSs such as WGS84), if the spherical geometry engine (s2) is enabled, as it is in **sf** by default (see section \@ref(s2)).
```

PR #922, opened 2023-02-10
Comment out failing code, fix 921

Robinlovelace:


jannes-m (approved):
thanks for the quick-fix, @Robinlovelace, then let's leave it at that until the mlr3 code works properly again

PR #928, opened 2023-02-25
Clarify how plot() works with sf objects

### LINKED ISSUE #924: Note on 2.2.1 world plot example
For context, im new to geocomp and am working my way through this great resource and am offering a perspective as a set of "fresh to geocomp" eyes.

Just a small suggestion, this issue refers to where the section references using the base r plotting function to plot the "world" dataset. While the link text directs the user to the terra::plot function, the section text states "sf objects can be plotted quickly with the base R function" which could be mistaken for using the base r::plot function instead of the terra::plot function. When I was working through the example, I found myself using the base plot function in place of terra and it may be worth explicitly stating that the plot function associated with sf is overwritten/expanded by the terra plot function.  

Also, I really appreciate your work! It has been really helpful to have such a thorough guide as a resource as someone whose transitioning from using ESRI ArcPro for geospatial to R!

Robinlovelace:
I think this will fix #924

### DIFF
--- a/02-spatial-data.Rmd
+++ b/02-spatial-data.Rmd
@@ -202,8 +202,9 @@ names(world)
 
 The contents of this `geom` column give `sf` objects their spatial powers: `world$geom` is a '[list column](https://adv-r.hadley.nz/vectors-chap.html#list-columns)' that contains all the coordinates of the country polygons.
 \index{list column}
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Merging now as fixes the issue.

PR #933, opened 2023-03-03
change grass82 to grass-stable

babayoshihiko:
Use grass-stable instead of version specific link


Nowosad:
Thank you @babayoshihiko! Great improvement

PR #934, opened 2023-03-03
Seems select does not need dplyr prefix?

babayoshihiko:
It seems a recent commit removed "tidyr::" prefix from select(), hence remove it in  this line as well.

There still are "tidyr::" in line 592, 601, 791.

Additionally, removed an unnecessary blank line.

### DIFF
--- a/03-attribute-operations.Rmd
+++ b/03-attribute-operations.Rmd
@@ -414,7 +414,7 @@ The following command calculates population density (with `mutate()`), arranges 
 ```{r 03-attribute-operations-30}
 world_agg5 = world |> 
   st_drop_geometry() |>                      # drop the geometry for speed
-  dplyr::select(pop, continent, area_km2) |> # subset the columns of interest  
+  select(pop, continent, area_km2) |> # subset the columns of interest  
   group_by(continent) |>                     # group by continent and summarize:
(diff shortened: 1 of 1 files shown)


Robinlovelace:
Agreed. We used the namespace for `select()` specifically to differentiate it from `raster::select()`. Now that we've switched to `terra` it's now longer an issue. 

https://rdrr.io/cran/raster/man/select.html

vs

`sel()` in **terra**. https://rspatial.github.io/terra/reference/select.html

03-attribute-operations.Rmd
```diff
@@ -414,7 +414,7 @@ The following command calculates population density (with `mutate()`), arranges
 ```{r 03-attribute-operations-30}
 world_agg5 = world |> 
   st_drop_geometry() |>                      # drop the geometry for speed
-  dplyr::select(pop, continent, area_km2) |> # subset the columns of interest  
+  select(pop, continent, area_km2) |> # subset the columns of interest
```
Robinlovelace, line 417: :+1:

Robinlovelace (approved):

03-attribute-operations.Rmd
Robinlovelace: Does that appear anywhere else in the book? Good to merge I think!

PR #936, opened 2023-03-04
Replace dplyr::select with select throughout

### LINKED ISSUE #935: dplyr::select() contains namespace in most uses
As shown here `dplyr::select()` is used more frequently than just `select()`: https://github.com/geocompx/geocompr/search?q=select

It's probably worth completing the job started in https://github.com/geocompx/geocompr/pull/934 for all instances for consistency, now that we're switched away from **raster**.

Robinlovelace:
close #935

### DIFF
--- a/09-mapping.Rmd
+++ b/09-mapping.Rmd
@@ -1002,7 +1002,7 @@ Before we cover mapping packages for plotting a specific type of map (in the nex
 ```{r map-gpkg, echo=FALSE, message=FALSE, warning=FALSE}
 # code/09-map-pkgs.R
 gpkg_df = readr::read_csv("extdata/generic_map_pkgs.csv")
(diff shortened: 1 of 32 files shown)


Robinlovelace:
Putting this in as a PR in case there are instances where **dplyr** is not yet loaded. We'll find out for the core parts of the book with CI. This finishes the job started in #934

Nowosad (approved):

Nowosad:
:+1

jannes-m:
For clarification purposes I used  package prefixes for all functions in chapter 12 but maybe this was not necessary for tidyverse functions. But then we should remove tidyverse prefixes from all functions.

Robinlovelace:
> For clarification purposes I used package prefixes for all functions in chapter 12 but maybe this was not necessary for tidyverse functions. But then we should remove tidyverse prefixes from all functions.

Agreed, will make the code more concise.

Issue #879, opened 2022-10-08
Add benchmarking results


Robinlovelace:
I'm curious: how does hardware affect book build times. I'm thinking we could use functions in **benchmarkme** to pull out CPU and other hardware/system details and add a file, e.g. called benchmark.csv reporting times taken to build the book on different platforms. A single script, e.g. called benchmark.R, could orchestrate the build and save the results. Just an idea!

Robinlovelace:
Finally taking a look at this... And running the build in a devcontainer.

PR #947, opened 2023-04-08
Add benchmarking code and initial CSV (#879)

Robinlovelace:


Nowosad (approved):

Nowosad:
Looks good! 👍🏻

Robinlovelace:
Thanks Jakub. I was thinking of adding a loop to benchmark each chapter and save each chapter result in a new column. I think that's worth it. We can do that as a post merge fix.

PR #957, opened 2023-04-14
Fix Docker command

Robinlovelace:
See https://rocker-project.org/images/versioned/rstudio.html


jannes-m (approved):
it doesn't matter if you use `true` or `True` but if `true` is the rocker recommendation, let's go with that

Robinlovelace:
Aha OK, didn't realise it can be `True` or `true` with Docker. Mild preference for latter but now see: this is style not a fix!

PR #964, opened 2023-05-10
mini-review of conclusions chapter

jannes-m:


16-synthesis.Rmd
```diff
@@ -57,9 +57,9 @@ identical(nz_name1$Name, nz_name2$Name) # check results
 
 This raises the question: which to use?
 The answer is: it depends.
-Each approach has advantages: the pipe syntax is popular and intuitive, while base R\index{R!base} tends to stable, well-known, and has minimal dependencies.
-Choosing between them is therefore largely a matter of preference.
-Our advice is to try both and see which you prefer.
+Each approach has advantages: base R\index{R!base} tends to be stable, well-known, and has minimal dependencies which is why it is often preferred for software (package) development.
```
jannes-m, line 60: I wouldn't mention the base pipe as an advantage for using the tidyverse since you can use it both with base R and the tidyverse (e.g., if you must, you can write ` nz |> `[`(_$Name)`). I guess this was just a leftover argument before switching from the magrittr pipe to the newly introduced base pipe. In any case, I have made here the argument that base R is often preferred for software development while the tidyverse is often preferred for interactive programming. But feel free to change.

jannes-m:
In general, I wonder if we still should mention the **sp** package since it will be retired this year. Well, maybe we should since so many packages still depend on it but I suggest to add to the conclusions  that **sf** will officially replace **sp** in 2023.

16-synthesis.Rmd
```diff
@@ -124,7 +124,7 @@ These resources include @zuur_mixed_2009, @zuur_beginners_2017 which focus on ec
 [*R for Geographic Data Science*](https://sdesabbata.github.io/r-for-geographic-data-science/) provides an introduction to R for geographic data science and modelling.
 
 We have largely omitted geocomputation on 'big data' by which we mean datasets that do not fit on a high-spec laptop.
-This decision is justified by the fact that the majority of geographic datasets that are needed for common research or policy applications *do* fit on consumer hardware.
+This decision is justified by the fact (that aside from large-scale fine-resolution remote sensing data) the majority of geographic datasets that are needed for common research or policy applications *do* fit on consumer hardware.
```
jannes-m, line 127: Just because we omit largely remote sensing data does not mean that it is also commonly neglected for geographic data analysis -:) Think e.g., of **stars** and OpenEO.

Nowosad:
> In general, I wonder if we still should mention the **sp** package since it will be retired this year. Well, maybe we should since so many packages still depend on it but I suggest to add to the conclusions that **sf** will officially replace **sp** in 2023.

sp is not going to be retired (rgdal, rgeos, and maptools will) soon.

16-synthesis.Rmd
Robinlovelace: Yes, that was a leftover argument. Good point Jannes and :+1: to the change.

Robinlovelace:
Ah just saw Jakub made the same point..

16-synthesis.Rmd
Robinlovelace: Agreed. Minor tweak I suggest is to simplify the adjectives and remove the brackets:

```
This decision is justified by the fact that the majority of geographic datasets that are needed for common research or policy applications *do* fit on consumer hardware, large high resolution remote sensing datasets being a notable exception.
```

16-synthesis.Rmd
```diff
@@ -108,23 +108,23 @@ length(spdeps) # 463
 
 ## Gaps and overlaps {#gaps}
 
-Geocomputation is a big area, so there inevitably gaps in this book.
+Geocomputation is a big area, so there are inevitably gaps in this book.
```
Robinlovelace, line 111: Definite typo fix, sorry there are many and many thanks for the fixes to this and more!

jannes-m: You are very much welcome and no problem at all.

16-synthesis.Rmd
jannes-m: Excellent, I'll write it like this.

jannes-m:
Thanks, Robin, for rewriting the Conclusions. Looks good and very much encourages the open-source approach +1

16-synthesis.Rmd
```diff
@@ -154,20 +155,20 @@ Trying each of the following steps --- skipping steps already taken --- provides
...
 You can also make online searches more effective by including additional detail:
 <!-- To provide a concrete example, imagine you want to know how to use custom symbols in an interactive map. -->
 
-- Use quote marks to maximise the chances that 'hits' relate to the exact issue you're having by reducing the number of results returned. For example, if you try and fail to save a GeoJSON file in a location that already exists, you will get an error containing the message "GDAL Error 6: DeleteLayer() not supported by this dataset". A specific search query such as `"GDAL Error 6" sf` is more likely to yield a solution than searching for `GDAL Error 6` without the quote marks.
+- Use quote marks to maximize the chances that 'hits' relate to the exact issue you're having by reducing the number of results returned. For example, if you try and fail to save a GeoJSON file in a location that already exists, you will get an error containing the message "GDAL Error 6: DeleteLayer() not supported by this dataset". A specific search query such as `"GDAL Error 6" sf` is more likely to yield a solution than searching for `GDAL Error 6` without the quote marks.
```
Robinlovelace, line 171: :+1:

Robinlovelace (approved):
This is good to go. Feel free to merge whenever ready @jannes-m. Thanks!

Issue #965, opened 2023-05-12
Check dependency checks 


Robinlovelace:
The check for SAGA here always returns false according to comments on https://github.com/geocompx/geocompr/commit/3ed05da71d65915d790f8600fd832eaa361b63c2:

https://github.com/geocompx/geocompr/blob/3ed05da71d65915d790f8600fd832eaa361b63c2/10-gis.Rmd#L24-L37

Related: https://github.com/geocompx/geocompr/commit/0a58256f5230359fce0cb30c7ed2811f4e7bf582 and https://github.com/geocompx/docker/issues/37

Robinlovelace:
Checking this by running the following line:

https://github.com/geocompx/geocompr/blob/f5a5c4e69f7a152f547bbca5333e92b45149ce2d/10-gis.Rmd#L140

Robinlovelace:
That generates this: 

![image](https://github.com/geocompx/geocompr/assets/1825120/384cbec8-490f-4a56-b6cb-204368eebf53)

Robinlovelace:
Issue with the container: 

``` r
qgisprocess::qgis_providers()
#> Attempting to load the cache ... Success!
#> # A tibble: 4 × 3
#>   provider provider_title    algorithm_count
#>   <chr>    <chr>                       <int>
#> 1 gdal     GDAL                           56
#> 2 qgis     QGIS                           50
#> 3 3d       QGIS (3D)                       1
#> 4 native   QGIS (native c++)             242
```

<sup>Created on 2023-05-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Robinlovelace:
So more motivation to address this: https://github.com/geocompx/docker/issues/37

Can you take a look please Jannes?

jannes-m:
Yes, let me do it 😊

jannes-m:
but I just checked it is not an issue with the container, in essence we have to use `qgis_plugins()` for checking and later on in the text (before calling `qgis_providers()`) we have to enable the plugins. But I will add this and an appropriate description in the text. I hope to do so this evening/night.

Robinlovelace:
Thanks Jannes, makes sense. I was getting confused between `qgis_plugins()` and `qgis_providers()` I think. Updated reprex showing that SAGA and GRASS are in there:

``` r
# Only run code if qgis is installed
 has_qgis_plugins = FALSE 
 # Check to see that the qgisprocess package is installed: 
[...]

PR #966, opened 2023-05-12
Qgis plugins

jannes-m:
addresses #965 and  replaces the .sdat/.tif hack with the `qgisprocess.tmp_raster_ext` option.

### DIFF
--- a/10-gis.Rmd
+++ b/10-gis.Rmd
@@ -173,16 +173,8 @@ You can do so from within QGIS with the [Plugin Manager](https://docs.qgis.org/3
 
 ```{r providers, eval=has_qgis_plugins}
 qgis_providers()
-#> # A tibble: 6 × 3
-#>   provider provider_title    algorithm_count
-#>   <chr>    <chr>                       <int>
-#> 1 gdal     GDAL                           56
-#> 2 grass7   GRASS                         304
-#> 3 qgis     QGIS                           50
-#> 4 3d       QGIS (3D)                       1
-#> 5 native   QGIS (native c++)             242
-#> 6 sagang   SAGA Next Gen                 509
 ```
+
 The output table affirms that we can use QGIS geoalgorithms (`native`, `qgis`, `3d`) and external ones from the third-party providers GDAL, SAGA and GRASS through the QGIS interface.
 
 We are now ready for some geocomputation with QGIS and friends, from within R!
@@ -263,19 +255,6 @@ The following command returns a data frame with each row representing an argumen
 alg = "native:union"
 union_arguments = qgis_arguments(alg)
 union_arguments
-#> # A tibble: 5 × 6
-#>   name    description qgis_type default_value available_values acceptable_values
-#>   <chr>   <chr>       <chr>     <list>        <list>           <list>           
-#> 1 INPUT   Input layer source    <NULL>        <NULL>           <chr [1]>        
-#> 2 OVERLAY Overlay la… source    <NULL>        <NULL>           <chr [1]>        
-#> 3 OVERLA… Overlay fi… string    <NULL>        <NULL>           <chr [3]>        
-#> 4 OUTPUT  Union       sink      <NULL>        <NULL>           <chr [1]>        
-#> 5 GRID_S… Grid size   number    <NULL>        <NULL>           <chr [3]>  
-
-#> [[1]]
-#> [1] "A numeric value"                                                                                 
-#> [2] "field:FIELD_NAME to use a data defined value taken from the FIELD_NAME field"                    
-#> [3] "expression:SOME EXPRESSION to use a data defined value calculated using a custom QGIS expression"
 ```
 
 The arguments, contained in `union_arguments$name`, are `INPUT`, `OVERLAY`, `OVERLAY_FIELDS_PREFIX`, and `OUTPUT`.
@@ -296,7 +275,6 @@ union = qgis_run_algorithm(alg,
   INPUT = incongr_wgs, OVERLAY = aggzone_wgs
 )
 union
-#>  $ OUTPUT: 'qgis_outputVector' chr "/tmp/...gpkg"
 ```
 
 Running the above line of code will save our two input objects into temporary .gpkg files, run the selected algorithm on them, and return a temporary .gpkg file as the output.
@@ -337,14 +315,7 @@ For this example, let's focus on just a few arguments, however, we encourage you
 ```{r, eval=FALSE}
 qgis_arguments("grass7:v.clean") |>
   select(name, description) |>
-  dplyr::slice_head(n = 4)
-#> # A tibble: 4 × 2
-#>   name      description
-#>   <chr>     <chr>
-#> 1 input     Layer to clean
-#> 2 type      Input feature type
-#> 3 tool      Cleaning tool
-#> 4 threshold Threshold (comma separated for each tool)
+  slice_head(n = 4)
 ```
 
 The main argument for this algorithm is `input` -- our vector object.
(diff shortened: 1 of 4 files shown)


10-gis.Rmd
```diff
@@ -26,13 +26,12 @@ library(gdalcubes)
...
-  # Check that QGIS is available to prevent errors in subsequent checks:
-  if (qgisprocess::has_qgis()) {
-    # Check that the plugins are installed (they will loaded later):
-    has_grass = qgisprocess::qgis_has_provider("grass7")
-    has_saga = qgisprocess::qgis_has_provider("sagang")
-    has_qgis_plugins = has_grass && has_saga
-  }
+  # Check that QGIS and required plugins are available to prevent errors in subsequent checks:
+  has_qgis_plugins = 
+    qgisprocess::has_qgis() &&
```
Robinlovelace, line 31: I suggest separating out the logic for QGIS being installed and for detecting plugins.

Pipes can be nice and help make data processing steps more intuitive. Here I do not find the piping helpful and suggest a simpler more verbose style would be beneficial to readers (if only readers who actually look at the source code ; )

That's my stylistic comment. In terms of the logic: we should stop instantly if QGIS is not installed. Hence previous versions had another if statement to prevent `qgis_plugins()` running when QGIS itself is not available.

10-gis.Rmd
```diff
@@ -153,34 +149,31 @@ There are a few possible solutions when the configuration fails: you can set `op
 <!-- link to the vignette https://github.com/r-spatial/qgisprocess/pull/31/files when it is online -->
 The above approaches can also be used when you have more than one QGIS installation and want to decide which one to use.
 
-Next, we can find which providers (meaning different software) are available on our computer.
+Next, we can find which plugins (meaning different software) are available on our computer.
 
-```{r providers, eval=has_qgis_plugins}
-qgis_providers()
-#> # A tibble: 6 × 3
```
Robinlovelace, line 160: Why remove the output?

Robinlovelace (changes requested):
This looks good and fixes #965 but removes output from the version of the book that we render. We could use the `qgis` image for rendering the main version but I think the disadvantages of that in terms of reproducibility of the book (most users will not have the right set-up with QGIS installed) and maintenance burden on developers (us) outweigh the benefits, if you see what I mean?

jannes-m:
Ah, my mistake, I thought we were going to use the qgis image. Though I have to admit I can't follow your reasoning why using it would be a burden for us because if users have not installed QGIS and required plugins, `has_qgis_plugins` will prevent the book from failing simply by not evaluating the corresponding code chunks. But ok, if you don't want to use the qgis image, I have to hardcode the output again (which of course is not very reproducible -:)).

10-gis.Rmd
jannes-m: Ok, this might be a matter of taste, and yes, we can get rid of the pipe, I don't mind. But I really try to avoid unnecessary `if` conditions, thinking about it, I would find it even more elegant to write the check like this:

```r
has_qgis_plugins = 
  # make sure that qgisprocess is installed
  "qgisprocess" %in% installed.packages() &&
  # make sure that QGIS is available
  qgisprocess::has_qgis() &&
  # make sure that all required plugins are available
    all(c("grassprovider", "processing_saga_nextgen") %in% 
    qgisprocess::qgis_plugins()$name)
```

In my opinion, this is quite readable, and avoids redundancies. So if any of these conditions are unmet, qgis chunks will be not evaluated.

jannes-m:
Use new qgisprocess function names, see https://github.com/r-spatial/qgisprocess/pull/153

Robinlovelace:
> Use new qgisprocess function names, see [r-spatial/qgisprocess#153](https://github.com/r-spatial/qgisprocess/pull/153)

That's a lot of changes! Well spotted..

Robinlovelace:
Regarding using the `qgis` image by default: we could and I can see the advantages from a reproducibility perspective. 

However, it's the maintenance burden of having the main build depend on a package that is not yet on CRAN which has huge dependency stack which is also and which may change. So my take is that yes: the pros of keeping the `latest` tag as the main book build engine outweigh the pros of shifting this to the QGIS image:

https://github.com/geocompx/geocompr/blob/f5a5c4e69f7a152f547bbca5333e92b45149ce2d/.github/workflows/main.yaml#L10

Robinlovelace:
One more idea: we could have two chunks

- One with `eval=has_qgis_plugins, echo=has_qgis_plugins`
- The other with `!has_qgis_plugins` for both

If that would work?

10-gis.Rmd
jannes-m: I have written it like this now (but now using `qgis_has_plugin()`) but if you don't like it, feel free to change it. It is really not that important and not worth arguing about.

jannes-m:
> One more idea: we could have two chunks
> 
>     * One with `eval=has_qgis_plugins, echo=has_qgis_plugins`
> 
>     * The other with `!has_qgis_plugins` for both
> 
> 
> If that would work?

This would probably work but then we would have duplicated code chunks. It's ok, I will re-add the hard-coded output. But not today anymore, need some sleep, last night was really rough since David needed to puke a few times -:)

10-gis.Rmd
```diff
@@ -23,16 +23,14 @@ library(gdalcubes)
...
-# Check to see that the qgisprocess package is installed:
-if ("qgisprocess" %in% installed.packages()) {
-  # Check that QGIS and required plugins are available to prevent errors in subsequent checks:
-  has_qgis_plugins = 
-    qgisprocess::has_qgis() && 
-    c("grassprovider", "processing_saga_nextgen") %in% 
-    qgisprocess::qgis_plugins()$name |>
-    all()
-}
+has_qgis_plugins =
```
Robinlovelace, line 26: My suggestion here: if `has_qgis()` is FALSE, there should not be any other `qgisprocess::*` functions called.

From a readability: it's worth creating separate objects stating if the environment has QGIS, grass, and saga, unless there are advantages of this more all-at-once-but-less-readable approach that I've not thought of.

10-gis.Rmd
```diff
@@ -145,9 +143,8 @@ library(qgisprocess)
 
 This package automatically tries to detect a QGIS installation and complains if it cannot find it.^[You can see details of the detection process with `qgis_configure()`.]
 There are a few possible solutions when the configuration fails: you can set `options(qgisprocess.path = "path/to/your_qgis_process")`, or set up the `R_QGISPROCESS_PATH` environment variable.
-<!--toDo:jn-->
-<!-- link to the vignette https://github.com/r-spatial/qgisprocess/pull/31/files when it is online -->
 The above approaches can also be used when you have more than one QGIS installation and want to decide which one to use.
+For more details, please refer to the **qgisprocess** ['getting started' vignette](https://r-spatial.github.io/qgisprocess/articles/qgisprocess.html).
```
Robinlovelace, line 147: :+1:

Robinlovelace (commented):
Looking good, minor comments on dependency checks...

Robinlovelace:
> This would probably work but then we would have duplicated code chunks. It's ok, I will re-add the hard-coded output. But not today anymore, need some sleep, last night was really rough since David needed to puke a few times -:)

:+1: :baby: 

Thanks Jannes!

Robinlovelace (approved):
Looking good.

Robinlovelace:
Will be interesting to see how this responds to using the new `qis` image that was built after https://github.com/geocompx/docker/pull/39 was finally merged. Long story, we now have more up-to-date (and pretty well supported by OSGEO/Rocker) upstream versions of SAGA etc in the image.

Robinlovelace:
This is looking good Jannes. Anything else we need to do before merging? Do you have any comments @Nowosad ?

jannes-m:
I still need to re-add the hard-coded output.

Robinlovelace:
> I still need to re-add the hard-coded output.

Sounds good. I would err on the side of merge all these improvements early, we can add stuff like that later down the line. Keen to keep momentum..

Robinlovelace:
Based on my comment

> Sounds good. I would err on the side of merge all these improvements early, we can add stuff like that later down the line. Keen to keep momentum..

And :+1: from @jannes-m, will merge now. Let's see what happens!

PR #972, opened 2023-05-23
Change polygon direction, close #970

### LINKED ISSUE #970: Point order in a polygon in chapter 11
@Robinlovelace, the convention is that the points building the polygon should be ordered counter-clockwise (see, for example, the last paragraph in https://r-spatial.org/book/03-Geometries.html#sec-valid).

That is not the case in chapter 11 (https://r.geocompx.org/algorithms.html), where the points are ordered clockwise. It is not a big problem, but I think it would be better to improve it (give "a good example" for others).

Robinlovelace:


Nowosad:
Hi @Robinlovelace -- these changes look good, but some more are needed:
1. FIGURE 11.1 should be updated
2. E1 in the Exercises: `poly_mat = cbind(x = c(0, 0, 9, 9, 0), y = c(0, 9, 9, 0, 0))`

Robinlovelace:
Well spotted re figure! Updated.

Nowosad (approved):

PR #997, opened 2023-11-29
add contraction hierarchies and traffic assignment in Transport chapter

vlarmet:
Hi !

Here is my little contribution.  
I tried to be as concise and clear as possible


Robinlovelace (approved):
This looks great to me, short and simple which is good because the book is already long. Will merge and make few post merge fixes!

PR #1000, opened 2023-12-04
Update 01-introduction.Rmd

tibbles-and-tribbles:
Typo


Robinlovelace (approved):
How did we not spot that?!

PR #1006, opened 2023-12-07
Update 02-spatial-data.Rmd

tibbles-and-tribbles:
Structure of MULTIPOINT geometry type requires additional parentheses, right?


Robinlovelace (commented):
I think the original is correct here.

tibbles-and-tribbles:
Thanks! I submitted an issue on this topic so I could explain myself better.

PR #1010, opened 2023-12-11
Update 03-attribute-operations.Rmd

tibbles-and-tribbles:
Spelling


Robinlovelace (approved):
You have eagle eyes. Many thanks for tidying up all our typos, surprised there are so many!

PR #1014, opened 2023-12-13
Update 03-attribute-operations.Rmd

tibbles-and-tribbles:
Removing "not" expresses the desired intent, yes?

### DIFF
--- a/03-attribute-operations.Rmd
+++ b/03-attribute-operations.Rmd
@@ -787,7 +787,7 @@ These and other typical raster processing operations are part of the map algebra
 ```{block 03-attribute-operations-65, type='rmdnote'}
 Some function names clash between packages (e.g., a function with the name `extract()` exist in both **terra** and **tidyr** packages).
 This may lead to unexpected results when loading packages in a different order.
-In addition to not loading packages by referring to functions verbosely (e.g., `tidyr::extract()`), another way to prevent function names clashes is by unloading the offending package with `detach()`.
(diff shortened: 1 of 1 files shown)


03-attribute-operations.Rmd
```diff
@@ -787,7 +787,7 @@ These and other typical raster processing operations are part of the map algebra
 ```{block 03-attribute-operations-65, type='rmdnote'}
 Some function names clash between packages (e.g., a function with the name `extract()` exist in both **terra** and **tidyr** packages).
 This may lead to unexpected results when loading packages in a different order.
-In addition to not loading packages by referring to functions verbosely (e.g., `tidyr::extract()`), another way to prevent function names clashes is by unloading the offending package with `detach()`.
+In addition to loading packages by referring to functions verbosely (e.g., `tidyr::extract()`), another way to prevent function name clashes is by unloading the offending package with `detach()`.
```
Robinlovelace, line 790:
```suggestion
In addition to calling functions verbosely with their full namespace (e.g., `tidyr::extract()`) to avoid attaching packages with `library()`, another way to prevent function name clashes is by unloading the offending package with `detach()`.
```

Make more sense?

Robinlovelace (changes requested):

tibbles-and-tribbles:
Yes, that makes more sense.

Robinlovelace (approved):

PR #1030, opened 2024-01-02
eblondel geographic metadata

Robinlovelace:
- text proposal for geographic metadata
- cleans metadata section
- Update 08-read-write-plot.Rmd
- Define FAIR

### DIFF
--- a/08-read-write-plot.Rmd
+++ b/08-read-write-plot.Rmd
@@ -714,11 +714,10 @@ Geographic metadata are a cornerstone of geographic information management, used
 They help make datasets FAIR (Findable, Accessible, Interoperable, Reusable) and are defined by the ISO/OGC standards, in particular the ISO 19115 standard and underlying schemas. 
 These standards are widely used within spatial data infrastructures, handled through metadata catalogs.
 
-Through R, the management of geographic metadata can be done with the **geometa** package that allows writing, reading and validating geographic metadata according to the ISO/OGC standard schemas. 
+Geographic metadata can be managed with **geometa**, a package that allows writing, reading and validating geographic metadata according to the ISO/OGC standards. 
 It already supports various international standards for geographic metadata information, such as the ISO 19110 (Feature catalogue), ISO 19115-1 and 19115-2 (geographic metadata for vector and gridded/imagery datasets), ISO 19119 (geographic metadata for service), and ISO 19136 (Geographic Markup Language) providing methods to read, validate and write geographic metadata from R using the ISO/TS 19139 (XML) technical specification.
-
-The creation of geographic metadata is based on an object-oriented approach.
-A base metadata object is created and its fields can be completed with dedicated methods:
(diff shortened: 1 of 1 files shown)


08-read-write-plot.Rmd
```diff
@@ -708,6 +708,39 @@ writeRaster(x = single_layer, filename = "my_raster.tif",
...
+md$save("my_metadata.xml")
+# read a metadata from an XML file
+md = readISO19139("my_metadata.xml")
+```
+
+The package comes with more [examples](https://github.com/eblondel/geometa/tree/master/inst/extdata/examples) and has been extended by packages such as **[geoflow](https://github.com/r-geoflow/geoflow)** to ease and automate the management of metadata
+
+In the field of standard geographic information management, the distinction between data and metadata is less clear.
+The Geography Markup Language (GML) standard and file format covers both data and metadata, for example. 
+The **geometa** package allows exporting GML (ISO 19136) objects from geometry objects modeled with **sf**.
```
Robinlovelace, line 741: Is there a link to this? May be worth including if so as not totally clear to me how this works @eblondel

08-read-write-plot.Rmd
```diff
@@ -708,6 +708,39 @@ writeRaster(x = single_layer, filename = "my_raster.tif",
...
+Such functionality allows use of geographic metadata (enabling the inclusion of metadata on detailed geographic and temporal extents, rather than simple bounding boxes, for example) and the provisioin of services that extend the GML standard (e.g., OGC Web Coverage Service).
```
Robinlovelace, line 742: Is my update correct here @eblondel, is it more about providing or interacting with services?

08-read-write-plot.Rmd
eblondel: I would not say the distinction is less clear. 
What would be the link for? do you would like a vignette for geometa focus on GML objects?

08-read-write-plot.Rmd
eblondel: not really on the last sentence

eblondel: GML is often known as data format (the main OGC format), but many GML elements are metadata-oriented, and are used in service descriptions (enabling interoperability when interrogating OWS services with ows4R, with geometa as decoder for these GML elements), but also in dataset metadata, for which some metadata building blocks expect GML elements, it's the case of extents (geographic, temporal, vertical). 

For example The usual way to provide a geographic extent is to provide a bounding box, but it's a rough simplification of extents, and in some cases (e.g. vessel transects, routes network, etc) you want to provide a finer extent, e.g. with simplified linear geometries. All this can be provided in metadata, and this is where from the ISO 19115 you jump into the GML.
With **geometa**, we have a generic encoder/decoder of all elements, including GML, and converters from **sf**, which let you easily move from R **sf** to GML representations part of ISO 19115 metadata.

PR #1037, opened 2024-01-10
Add code for front cover

BjnNowak:
Code for the front cover of the second edition

### DIFF
--- a/code/front_cover2.R
+++ b/code/front_cover2.R
@@ -14,103 +14,95 @@ library(tidyverse)
 library(od)
 
 # Set fonts
-font_add_google("Raleway","ral")
+font_add_google("Raleway", "ral")
 showtext_auto()
 
 # Set size
 gg_record(
-  dir = file.path(tempdir(),"recording"), 
-  device = "png", 
-  width = 15.6, # I chose the size of the 1st edition book, so the image created here may cover all of the first page 
-  height = 23.4, 
-  units = "cm", # may also be exported as svg
-  dpi = 600 
+  dir = file.path(tempdir(), "recording"),
+  device = "png",
+  width = 15.6,
+  # I chose the size of the 1st edition book, so the image created here may cover all of the first page
+  height = 23.4,
+  units = "cm",
+  # may also be exported as svg
+  dpi = 600
 )
 
 # Cover is made of two different type of maps (1 "buffer" map and 1 "dorling" cartogram) that will be created one after another
-
 
 #############################
 # Creation of the first map #
 #############################
 
-# Let's start with the 1st map: buffer area around capitals in Africa
+# Let"s start with the 1st map: buffer area around capitals in Africa
 
 # Loading data
 ##############
 # Vector files of world borders loaded from {rnaturalearth}
-mp_with_countries<-ne_countries(scale = 50, type = "countries", returnclass = "sf")%>%
-  st_transform(crs="+proj=robin")
+mp_with_countries = ne_countries(scale = 50, type = "countries", returnclass = "sf") |>
+  st_transform(crs = "+proj=robin")
 
 # Populated places layer has been downloaded from Natural Earth:
 # https://www.naturalearthdata.com/downloads/10m-cultural-vectors/
 # For script reproducibility purposes, I uploaded a subset on my github:
-cp<-read_sf('https://github.com/BjnNowak/geocomputation/raw/main/data/ne_10m_populated_places.gpkg')%>%
+cp = read_sf("https://github.com/BjnNowak/geocomputation/raw/main/data/ne_10m_populated_places.gpkg") |>
   # Select only administrative capitals
-  filter(ADM0CAP==1)%>%
+  filter(ADM0CAP == 1) |>
   # Reproject to Robinson
-  st_transform(crs="+proj=robin")%>%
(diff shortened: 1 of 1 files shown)


Nowosad:
I like the new version.

I plan to look at the code it tomorrow and update it to the geocompr style.

Nowosad:
Dear @BjnNowak @Robinlovelace @jannes-m -- I just cleaned and updated the code. I also have two questions:

1. Do you prefer the "three palettes" version (top) or the "one palette" version (bottom) of the cover?
2. Do you prefer keeping the names of African cities or removing it from the final cover?

test-map
test-map-uc (1)

jannes-m:
Excellent work from all of you, looks fantastic!!! I would go with the three-palette version and actually I would keep the names to make the Africa map less 'empty'.

BjnNowak:
On my side:
- I do prefer the "three palettes" version (top) because it allows to see more see the difference between the three types of map
- I think leaving the names illustrates an additional mapping technique.

eblondel:
Hello, jumping into this, and congrats for this nice cover. 
Working daily with international boundaries, i'm wondering where you got the country & territories boundaries layer. Since we deal with a book cover on geospatial data, and that you display countries & territories boundaries, they are some boundary issues (e.g. Western Sahara, Abyei region, boundary Egypt/North Sudan) that might deserve to be addressed. In case you would like to fix these, and if I can support, let me know. Cheers

BjnNowak:
Country boundaries are from Natural Earth. Issue may be related to Natural Earth position on this subject, as stated on their website : "'Natural Earth shows boundaries by default according to who controls the territory, versus de jure." ?

Nowosad:
Regarding borders, do you have any suggestions on what could be done here @eblondel ?

Robinlovelace:
> On my side:
> 
>     * I do prefer the "three palettes" version (top) because it allows to see more see the difference between the three types of map
> 
>     * I think leaving the names illustrates an additional mapping technique.

:+1: to that, my preferences also.

Robinlovelace:
Regarding the country borders, what are the specific issues? I think it's clear that the purpose here is to visualise some data, not to endorse any particular boundary. But if there's a quick way to improve the map as Jakub suggests let us know.

BjnNowak:
One of my objectives with the proposal was to use as many "directly" accessible datasets as possible. Hence the TidyTuesday dataset for the cartogram and {rnaturalearth} for country boundaries. But if those boundaries are an issue, a quick fix would be to just change the source of the boundary files 'mp_with_countries' (L43 of Jakub's version)

@eblondel do you have any url to access a 'clean' vector file of country boundaries, with at least one column for country name and one for continent in the table of attribute?

eblondel:
@BjnNowak They are more official sources of truth, although I understand the logic that you followed to get data using a data-oriented R package and the choice of rnaturalearth.

To reply to @Robinlovelace the matter is not to endorse any boundary obviously, but you publish a book on geospatial data, and the cover uses country/territory boundaries. I believe it would give more credit to this book if you display accurate boundaries. Again it's just a suggestion.

As datasources:
- This is from the World Bank: https://datacatalog.worldbank.org/search/dataset/0038272/World-Bank-Official-Boundaries it's probably based on the UN cartographic section datasets.
- Take a look at this https://data.apps.fao.org/catalog/iso/6aecaf08-c7d0-4e16-8778-b8dddae9da14 We have an equivalent as OGC web-service (as WFS) maintained at UN-FAO Fisheries & Aquaculture divsion, that dispatchs the countries & territories boundaries defined by the UN cartographic section (they don't provide the web-services to download directly the data from a client).

If you want to quickly compare boundaries between natural earth and UN source, you can see these views:
- https://geoservices.un.org/Html5Viewer/index.html?viewer=clearmap
- https://www.fao.org/fishery/geoserver/wms?service=WMS&version=1.1.0&request=GetMap&layers=UN_intbnd_layergroup&bbox=-180.0%2C-90.0%2C180.0%2C90.0&width=768&height=384&srs=EPSG%3A4326&styles=&format=application/openlayers

You can use this code:

```
require(ows4R)
WFS = WFSClient$new(url = "https://www.fao.org/fishery/geoserver/wfs", serviceVersion = "1.0.0", logger = "INFO")
boundaries = WFS$getFeatures("fifao:UN_intbnd_legacy")
```
Attributes give you countries/territories involved in the boundary (as ISO 3 code), and you also have the type of boundaries, in case you want to display it in different styles. I don't think it's really needed for your cover, but at least you will get more accurate boundaries.
This layer is a relatively low resolution, probably enough for the cover. There is a higher resolution, but it might increase substantially the file size of your cover.

Cheers

BjnNowak:
Thx for the file @eblondel 

I kept Natural Earth vector for the continent selection, but used the new boundary files in the plot. See result below. [Script](https://github.com/BjnNowak/geocomputation/blob/main/script/geocomputation_cover.R) still in my repo (including Jakub's latest modification, I could make a new PR later if it's ok for you ? 

cover_un_boundaries

Robinlovelace:
Great work @BjnNowak we have our new cover! Let us know if you'd like to write a blog post about the process, see here for info: https://github.com/geocompx/geocompx.org/blob/main/post/2023/map-cover-competition/index.qmd We should do some kind of write-up.

PR #1039, opened 2024-01-11
Update 05-geometry-operations.Rmd

tibbles-and-tribbles:
typos


Robinlovelace (approved):
Agreed, I'm not a fan of -- at all, --- or , usually better. Thank you!

PR #1052, opened 2024-01-29
Chapter 13 (transportation): some fixes for the routing paragraph

florisvdh:
Unsure about the 2nd commit, though.

### DIFF
--- a/13-transport.Rmd
+++ b/13-transport.Rmd
@@ -495,7 +495,7 @@ Routing services based on open source routing engines, such as OSRM's publicly a
 However, the fact that external routing services are hosted on a dedicated machine (usually funded by commercial company with incentives to generate accurate routes) can give them advantages, including:
 
 - Provision of routing services worldwide (or usually at least over a large region)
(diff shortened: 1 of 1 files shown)


13-transport.Rmd
```diff
@@ -552,7 +552,7 @@ Making the width of the routes proportional to the number of car journeys that c
 The code chunk below plots the desire lines and routes, resulting in Figure \@ref(fig:routes) which shows routes along which people drive short distances:[^13-transport-8]
```
florisvdh, line 552: > The code chunk below

Something needs a change here: either drop `echo = FALSE` for the below chunk, or don't refer to the chunk that generates the plot.

Robinlovelace: Thanks for the heads-up, will fix now.

Robinlovelace (approved):
These are definite typos, some pretty bad, what was I thinking/typing :see_no_evil: Will address the issue around line 552 after merging.

PR #1060, opened 2024-01-31
Update 07-reproj.Rmd

tibbles-and-tribbles:
grammar


Robinlovelace (approved):
Both could work but agree, commas are more conventional and appropriate here. I'd like to get your reasoning here @tibbles-and-tribbles but it's a general :+1: from me.

Robinlovelace:
:+1: good explanation, thanks for your attention to detail :monocle_face:

PR #1061, opened 2024-01-31
Update 07-reproj.Rmd

tibbles-and-tribbles:
typo/grammar

### DIFF
--- a/07-reproj.Rmd
+++ b/07-reproj.Rmd
@@ -543,7 +543,7 @@ The output is printed as multiple lines of text containing information about the
 st_crs(cycle_hire_osm)
 ```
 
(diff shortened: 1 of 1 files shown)


07-reproj.Rmd
```diff
@@ -543,7 +543,7 @@ The output is printed as multiple lines of text containing information about the
 st_crs(cycle_hire_osm)
 ```
 
-As we saw in Section \@ref(crs-setting), the main CRS components, `User input` and `wkt`, are printed as a single entity, the output of `st_crs()` is in fact a named list of class `crs` with two elements, single character strings named `input` and `wkt`, as shown in the output of the following code chunk:
+As we saw in Section \@ref(crs-setting), the main CRS components, `User input` and `wkt`, are printed as a single entity. The output of `st_crs()` is in fact a named list of class `crs` with two elements, single character strings named `input` and `wkt`, as shown in the output of the following code chunk:
```
Robinlovelace, line 546: :+1: better as 2 sentences.

Robinlovelace (approved):

PR #1073, opened 2024-02-12
Update 09-mapping.Rmd

tibbles-and-tribbles:
Removed "title" because Figure 9.10 does not show output for changing the title.


Nowosad:
👍🏻

PR #1079, opened 2024-02-23
Update _05-ex.Rmd

smkerr:
- Fix typo: question asks for points within 100 km of Canterbury, but answer is within 100 m
- Alternative fix: ask for meters rather than kilometers


Nowosad:
Thanks!

PR #1089, opened 2024-05-03
Update _404.Rmd

edzer:


Robinlovelace (approved):
Looks like a good catch.

Before:

image

After, let's see!

Robinlovelace:
Fixed :tada: 

image

PR #1101, opened 2024-06-12
Front cover reproducibility improvements

Robinlovelace:


Nowosad:
👍🏻 

(One possible change would be to use `sapply(pkgs, require, character.only = TRUE)` as it returns nicer output in this case)

Issue #1115, opened 2024-09-23
Proofreading changes


Robinlovelace:
- [x] Foreword (2nd Edition): RL
- [x] Preface: RL
- [x] 1 Introduction: RL, JM, JN
- [x] 2 Geographic data in R: RL
- [x] 3 Attribute data operations: RL
- [x] 4 Spatial data operations: RL
- [x] 5 Geometry operations: JM
- [x] 6 Raster-vector interactions: JN
- [x] 7 Reprojecting geographic data: JN
- [x] 8 Geographic data I/O: JN
- [x] 9 Making maps with R: JN
- [x] 10 Bridges to GIS software: JN
- [ ] 11 Scripts, algorithms and functions: JM
- [ ] 12 Statistical learning: JM
- [ ] 13 Transportation: RL
- [ ] 14 Geomarketing: JM
- [ ] 15 Ecology: JM
- [x] 16 Conclusion: RL, JM, JN
- [ ] References: RL, JM, JN

Nowosad:
Chapters 6, 7, and 8 are done: https://github.com/geocompx/geocompr/pull/1119, https://github.com/geocompx/geocompr/pull/1120, https://github.com/geocompx/geocompr/pull/1121. I plan to work on Chapters 9 and 10 tomorrow.

Nowosad:
There are two remaining issues:

1. Several times, the comment is "AU: Please move this line of text to the next page" -- is seems like a lost battle, as when we move one line, then another problematic lines could occur)
2. References also contain some comments from the proof reader

jannes-m:
I plan to start the proof-reading today in the evening. @Robinlovelace I would start with the chapters that are solely assigned to me. When I am done with those, I'll also have a look at jointly assigned chapters.

jannes-m:
The reviewer points out (ch5): 
> Is it OK that you are spelling "gray" with an "a" here but "grey" with an "e" in the code below? Please do a global search for the word you wish to change (if you decide to do so).

I guess he has a point here, if you agree I would replace all instances of grey (BE) by gray (AE).

jannes-m:
c5 in #1128

Robinlovelace:
> I guess he has a point here, if you agree I would replace all instances of grey (BE) by gray (AE).

Agreed.

jannes-m:
c12 is done in https://github.com/geocompx/geocompr/pull/1136

Robinlovelace:
[...]

PR #1116, opened 2024-09-23
Complete proofing updates to foreword

Robinlovelace:


Robinlovelace:
Let's review, tweak if necessary, and then merge this PR. Maybe we should split the proof edits into 3 PRs, one for each part, each led by one of us, as per #1115 ?

Robinlovelace:
Merging now to avoid conflicts later.

PR #1144, opened 2024-12-09
Update 02-spatial-data.Rmd

arunkhattri:
correct repo address for "spDataLarge"  for package installation


Robinlovelace (approved):
Good catch, thank you!

Nowosad:
I think that both should work (as the geocompr universe uses the spDataLarge package that is at my repo (nowosad/spDataLarge)), but I also like the use of "geocompr" name in the book more.

Robinlovelace:
Would "geocompx" be better?

PR #1145, opened 2024-12-14
remove duplicate 'or not'

jacobvjk:
The original text had a duplicate 'or not' at different places in the same sentence. This should be clearer.


Robinlovelace (approved):
Definite improvement, thanks.

PR #1146, opened 2024-12-28
Update definition of geocomputation

Robinlovelace:
CC @Nowosad this is more concise and clear. Minor tweak, too late for print version of v2 but better I think.


Robinlovelace:
Nudge @Nowosad

PR #1161, opened 2025-06-15
Update 04-spatial-operations.Rmd - Fig 4.10 refers to different code …

cortinah:
…chunk example

Fig 4.10 refers to the masking, not the subsetting code chunk example.  Removed reference and moved figure slightly down.


Robinlovelace (approved):
Looks good to me but would like @Nowosad and @jannes-m to review before merging.

jannes-m (approved):

Nowosad (approved):

